PR #16, opened 2021-03-30
conditionally turns raiden off and on

Tunji17:


client/storage/index.js
```diff
@@ -13,6 +13,7 @@ const {
     createChannel,
     makePayment,
   } = require('./payments')
+const defaultConfig = require('../../resources/defaultConfig.json');
```
jensendarren, line 16: @Tunji17 - each node has its own config that is merged with the default config and is available via `ctx.config` so there is no need to import the defaultConfig.json like this.

client/storage/index.js
```diff
@@ -398,20 +399,18 @@ class Storage {
...
                     };
                     link.validatePledge();
-
-                    // Uncomment after demo for payments
-
-                    // const provider = await link.provider
-                    // const checksumAddress = await this.ctx.web3bridge.toChecksumAddress(`0x${provider.id.split('#')[1]}`)
-                    // await makePayment(checksumAddress, 10) // todo: calculate amount using cost per kb for service provider
-
+                    if (defaultConfig.raiden.switch === 'ON') {
```
jensendarren, line 402: @Tunji17 - see above about the ctx object. Its possible to access a nodes config like so `this.ctx.config` ... and then any of the params that are in the `defaultConfig`.

client/storage/machines/storage_link_machine.js
```diff
@@ -75,9 +77,12 @@ exports.createStateMachine = function createStateMachine(link, chunk) {
               ctx.client.deployerProgress.update(`chunk_${chunk.id}`, 0, link.state)
               return storage.SEND_STORE_CHUNK_REQUEST(chunk, link)
             },
-            onDone: {
+            onDone: [{
               target: 'creating_payment_channel',
-            },
+              cond: () => defaultConfig.raiden.switch === 'ON',
```
jensendarren, line 82: @Tunji17 same here. Can use `ctx.config....`

resources/defaultConfig.json
```diff
@@ -79,5 +79,8 @@
   },
   "simulation_delay": 10000,
   "hardcode_default_provider": "http://127.0.0.1:12345/#c01011611e3501c6b3f6dc4b6d3fe644d21ab301",
-  "token_address": "0xbf9be54Df2001E6Bd044cED0E508d936A9d38b1D"
+  "token_address": "0xbf9be54Df2001E6Bd044cED0E508d936A9d38b1D",
+  "raiden" : {
```
jensendarren, line 83: @Tunji17 - I would make a generic property, like `payments`.

resources/defaultConfig.json
```diff
@@ -79,5 +79,8 @@
...
+    "switch": "OFF"
```
jensendarren, line 84: @Tunji17 - I recommend making this a boolean, like `"enabled": false`. That way it can be applied in the node like so: `if (this.ctx.config.payments.enabled) { ... }`

jensendarren (commented):
@Tunji17 - Thank you for these changes. I've made a review and I'll go ahead and make the changes to the code myself and commit to this branch and then merge it into master. I just review so that you are aware of what is possible wrt the node configuration. Many thanks!

Tunji17:
@jensendarren Yeah, thanks for the review and update.

PR #23, opened 2021-04-29
Encrypt email data

Tunji17:


client/proxy/index.js
```diff
@@ -267,6 +276,75 @@ class ZProxy {
...
+                //persist encrypted message to storage layer and obtain hash
+                const cache_dir = path.join(this.ctx.datadir, this.config.cache_path);
+                this.ctx.utils.makeSurePathExists(cache_dir);
+                const tmpPostDataFilePath = path.join(cache_dir, this.ctx.utils.hashFnHex(encryptedMessage));
+                fs.writeFileSync(tmpPostDataFilePath, encryptedMessage);
+                let uploaded = await this.ctx.client.storage.putFile(tmpPostDataFilePath);
+                let encryptedMessageHash = uploaded.id;
+                console.log('encryptedMessageHash', encryptedMessageHash);
+
+                const emailContractAddress = await this.ctx.web3bridge.getKeyValue('email', 'zweb/contracts/address/Email')
```
jensendarren, line 308: @Tunji17 - Note that the 'Email' app and the 'Email.sol' Smart Contract are part of an Example application that would be deployed separately. So the key used `email` is referring to the deployed contract as part of a custom deployment and not part of the core Point Network node. It just means that if the contract name is changed in the example email ZHTML app then the `getKeyValue` function would return `undefined` for the `emailContractAddress`. This is just a note FYI as, for now, it can remain as it is with a hardcoded key here.

client/proxy/index.js
```diff
@@ -267,6 +276,75 @@ class ZProxy {
...
+                const emailContractAbi = await this.get_from_ikv('email', 'zweb/contracts/abi/Email');
+                const { abi } = JSON.parse(emailContractAbi)
+                const emailContract = new this.web3.eth.Contract(abi, emailContractAddress)
+                await emailContract.methods.send(recipient, encryptedMessageHash, encryptedSymmetricKey).call()
```
jensendarren, line 312: @Tunji17 Also note there is a helper function in `web3bridge` that can be used to invoke a Smart Contract method which is called like this `await this.ctx.web3bridge.sendContract(host, contractName, methodName, params);`. I suggest to use that helper instead of building out the contract instance like this.

jensendarren (commented):
@Tunji17 I've tested the application in this branch and I notice that the client side email notifications no longer appear to work. To reproduce this try sending an email from Node 1 to Node 2 and have the email applications open in two browser windows (same as in the demo video). You should notice that the notifications no longer work. You will also see in the browser console that the email counter is always showing a count of zero even after invoking send from the compose form. 

```
emails = await email.getPastEvents("SendEmail", {fromBlock: 0, toBlock: 'latest'}) // emails is always empty
```

I notice this PR is tagged `work in progress` so I thought I would bring this to your attention as something that needs to be looked into before a merge can be applied.

jensendarren (approved):

client/encryptUtils.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+  }
+
+module.exports.decryptCipherTextAndKey = async (host, cypertext, encryptedSymmetricObj, privateKey) => {
+    const decryptHostNameHash = crypto.createHash('sha256');
+    decryptHostNameHash.update(host);
+    const symmetricObj = await eccrypto.decrypt(Buffer.from(privateKey, 'hex'), encryptedSymmetricObj)
+    const [, hostNameHash, symmetricKey, iv] = symmetricObj.toString().split('|')
+    if (decryptHostNameHash.digest('hex') !== hostNameHash){
+        return {plaintext: null, hostNameHash: null, symmetricKey: null, iv:null}
```
tankakatan, line 39: I would better throw an error here. It will be easier to figure out what is actually going on.

client/zweb/renderer/index.js
```diff
@@ -103,6 +123,50 @@ class Renderer {
...
+                            message: decryptedResult.plaintext.toString(),
+                        }
+                    }
+                }))
+                return emails;
+            });
+            Twig.exports.extendFunction("decrypt_one_mail", async(host, owner, encryptedData) => {
+                const { privateKey } = this.ctx.wallet.config;
+                if (owner === encryptedData['1']) {
+                    const encryptedSymmetricKey = JSON.parse(encryptedData['3'])
```
tankakatan, line 153: The body of this `if` block looks very similar to the email decryption happening in the `retrieve_mail` function above (lines 134-145). I think it is a good idea to move it to a separate function in order to tolerate possible changes in `decryptCipherTextAndKey` interface.

tankakatan (approved):

tankakatan:
I have pulled the latest changes and tested the workflow. Everything works fine. The only minor thing I noticed is that right after an email successfully sent the app user gets signed out. I'm not sure it can be easily fixed due to the multiple redirects taking place on `encrypt and send` button click.

Tunji17:
> I have pulled the latest changes and tested the workflow. Everything works fine. The only minor thing I noticed is that right after an email is successfully sent the app user gets signed out. I'm not sure it can be easily fixed due to the multiple redirects taking place on `encrypt and send` button click.

Yeah, I think that's something we can resolve later on, also I have made the changes you suggested to the PR, It should be good to merge

jensendarren:
@Tunji17 I notice that after sending an email the user is no longer authenticated. In the `email-example-app` branch the user remains authenticated after sending the email. It appears to be loosing the session entry for `_session_auth` when posting to `_encrypt_send`.

jensendarren:
@Tunji17 Literally just saw @tankakatan comment about the authentication issue too! I think we can merge this into the app branch but I would suggest we fix the authentication issue as soon as possible as it will be better for the demo. Thank you!

PR #30, opened 2021-05-13
adds contract for store and contract

Tunji17:
> adds contract for store and contract
> starts integrating with the store dapp


jensendarren (changes requested):
@Tunji17 - I just merged @tankakatan PR into the `store-example-app` branch. As you know that contains a React JS client application now. Can you merge the `store-example-app` branch into your feature branch and then consider what changes you need to make to the app views to integrate the Smart Contracts with the React JS version. 

Additionally, since the React JS app requires an npm install and npm build step, I suggest adding the `@openzeppelin/contracts` dependency to the example Store sites `package.json` and referencing the contracts that way (unless that proves difficult for the deployment process, of course).

jensendarren (approved):

PR #31, opened 2021-05-13
Store example app frontend

tankakatan:
* Added `store.z` react typescript app
* Added `parcel` bundler scripts
* Added site deployment automation npm scripts
* Simplified site deployment workflow for sites that have `react: true` option in their `point.deploy.json` config
* Renderer now applies only to `zhtml` or `html` files


jensendarren (approved):
It looks good and works as expected locally on my computer running with and without Docker so I think it can be merged.

Tunji17 (approved):

PR #43, opened 2021-05-28
idea for using multiple versions of solc

Tunji17:
> adds function to fetch solidity file version
> checks for contract dependencies in contract and node modules folder
> injects dependencies and compile contract

### DIFF
--- a/client/zweb/deployer/index.js
+++ b/client/zweb/deployer/index.js
@@ -114,7 +114,7 @@ class Deployer {
         for(let contractName of contractNames) {
             let fileName = path.join(deployPath, 'contracts', contractName+'.sol');
             try {
-                await this.deployContract(target, contractName, fileName);
+                await this.deployContract(target, contractName, fileName, deployPath);
             } catch(e) {
                 this.ctx.log.error(e);
                 throw e;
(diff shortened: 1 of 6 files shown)


client/zweb/deployer/index.js
```diff
@@ -146,13 +146,35 @@ class Deployer {
         console.log('Deploy finished');
     }
 
-    async deployContract(target, contractName, fileName) {
+    static async getPragmaVersion(fileName){
```
jensendarren, line 149: @Tunji17 Just a suggestion... since the `deployContract` function reads the entire contents of the smart contract, perhaps its possible to simply pass the source directly to the `getPragmaVersion` function and then perform a regx on that string, something like this:

```
    static async getPragmaVersion(source){
        let regex = /pragma solidity [\^\~\>\<]?=?(?<version>[0-9\.]*);/;
        let found = null
        if (found = source.match(regex)) {
            return found.groups.version
        } else {
            throw new Error('Contract has no compiler version')
        }
    }
```

package.json
```diff
@@ -36,7 +36,8 @@
     "react-dom": "^17.0.1",
     "readline": "^1.3.0",
     "sanitize-html": "^2.3.2",
-    "solc": "^0.6.6",
+    "solc0_8": "npm:solc@^0.8",
+    "solc0_6": "npm:solc@^0.6",
```
jensendarren, line 40: @Tunji17 Can we also include compiler version 0.5.x and 0.7.x (where `x` is always the latest corresponding `PATCH` version). That way the node can support 4 versions out of the box.

example/twitter.z/contracts/Twitter.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.6.0;
+pragma solidity ^0.8.0;
```
jensendarren, line 2: @Tunji17 we can probably bump up all example sites contracts to `^0.8.0` in this branch.

jensendarren (commented):
@Tunji17 I've left some inline comments.

client/zweb/deployer/index.js
```diff
@@ -170,11 +192,14 @@ class Deployer {
...
-            switch (dependency) {
-                case contractName+'.sol':
-                    return {contents: fs.readFileSync(fileName, 'utf8')};
-                default:
-                    return {error: 'File not found'}
+            const dependencyLocalPath = path.join(deployPath, 'contracts', dependency)
+            const dependencyNodeModulesPath = path.join(deployPath, 'src/node_modules/', dependency)
+            if (fs.existsSync(dependencyLocalPath)) {
+                return {contents: fs.readFileSync(dependencyLocalPath, 'utf8')};
+            } else if (fs.existsSync(dependencyNodeModulesPath)){
```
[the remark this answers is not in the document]
Tunji17: Yes i have

client/zweb/deployer/index.js
```diff
@@ -146,13 +146,35 @@ class Deployer {
...
+        if (!version) throw new Error('Contract has no compiler version')
+    }
+
+    async deployContract(target, contractName, fileName, deployPath) {
         this.ctx.client.deployerProgress.update(fileName, 0, 'compiling')
+        const version = await this.constructor.getPragmaVersion(fileName);
+        const versionArray = version.split('.');
+
+        let SOLC_MAJOR_VERSION = versionArray[0]
+        let SOLC_MINOR_VERSION = versionArray[1] // TODO Change this based on the solidity pragma statement
```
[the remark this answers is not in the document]
Tunji17: sure

client/zweb/deployer/index.js
Tunji17: i'll test this out and get back to you on this

package.json
Tunji17: Yes i will include this

example/twitter.z/contracts/Twitter.sol
Tunji17: okay

client/zweb/deployer/index.js
jensendarren: OK thank you. What you have done works of course. In order to implement the function like this you would need to read the source of the contract in the `deployContract` before calling `getPragmaVersion` and store the source in a variable to then use in the `compileConfig` object. It should be pretty straightforward to change, but if it causes issues. then we can leave it.

Tunji17: Yeah I understand

jensendarren (approved):

PR #44, opened 2021-05-29
Introducing Directories

pointnetwork:


Tunji17:
I think this works as it's expected to, I noticed an error during compiling twitter.sol which I fixed in #45, it should be merged to this branch before it is finally merged into master

PR #52, opened 2021-06-19
Initializing eslint configuration file.

amherag:
For now, we have 2 rules: 4 spaces for indentation and semicolons are enforced.


jensendarren (changes requested):
Hi @amherag I think its worth installing `eslint` as a project dev dependency and then updating the package.json files too.

amherag:
I added `eslint` as a depdency.

jensendarren (approved):

PR #57, opened 2021-06-27
Fix #55: The localPath of all the files on the node should be set to the nodes storage layer.

### LINKED ISSUE #55: The `localPath` of all the files on the node should be set to the nodes storage layer.
**Originally described by @jensendarren**

### Summary

The `localPath` of all the files on the node should be set to the nodes storage layer. Currently it can be set **outside of the the storage layer** which is incorrect.

### Overview

I've noticed since moving to directories deployment, the files `localPath` can be set to the git working directory which results in being able to change files on the fly without having to run a subsequent point deploy. This is because the files `localPath`  is set to the _full path of the examples folder of the pointnetwork project_. What this effectively means is that it's only necessary to deploy the site once and then subsequent changes can be made directly in the the project files without having to redeploy!

### To replicate

1. Start a clean node without any sites deployed.
1. Open the Point Browser and **set the proxy to point to Node 2** (the _site owner_ node). 
1. Now navigate to http://twitter.z . You will see the error `500: ZProxy Error: Error: getRootDirectoryForDomain failed: key ::rootDir` returned empty as expected.
1. Deploy the twitter site as normal
1. Now reload the page and you will see the site load as expected
1. Now delete the twitter.z index.html file **in the project working directory** - not in the storage layer.
1. Open the site again and you will see an error that the file is missing and that the file path is set to the git repo working directory.

### Checking the Files Level DB

To view the entries in the Level DB you can use the API endpoint `/api/storage/files` which reveals the `localPath` is pointing to the project git working directory as described above. This should **not** be happening and the `localPath` should always be set to the designated storage layer directory of the node (usually `~/.point/testN` where `N` is the node number.

```
{
    "files": [
        {
            "id": "4966b3a9abda010b51069d49c9f1e146f3f499319a48cc6f29a2e052adc443e3",
            "localPath": "/Users/darren/workspace/pn/pointnetwork/example/hello.z/public/index.html",
            "size": 198,
            "redundancy": 1,
            "expires": 1622682265211,
            "autorenew": true,
            "chunkCount": 1
        },
}
```

amherag:
Please review if the place where I'm doing the re-assignment is the appropriate place.

According to what I understood after reviewing several parts of the source code, we use `File.localPath` and `Directory.localPath` to set the original sources of the files and directories being read, respectively. At the moment we want to upload a file via `storage.putFile`, we use `File.localPath` to `File.chunkify` the file, which is the expected behavior. However, after chunkifying the file, we want to `model.save` it to our LevelDB database. In this step, `File.localPath` needs to be set to the newly created chunk file, which is named `'chunk_id' + File.id`, where `File.id` is the `id` given to the file by `chunkify`, instead of using the original file's path name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ Should return the following result:
 ```json
 {
    "file": {
-       "localPath": "/Users/developer/.point/test1/data/client_storage_cache/file_45c1cc29130796f083425310aad7c10ed9c2a4cf7031e69dd40cd8551af65af2",
+       "originalPath": "/Users/developer/.point/test1/data/client_storage_cache/file_45c1cc29130796f083425310aad7c10ed9c2a4cf7031e69dd40cd8551af65af2",
        "dl_status": "ds99",
        "id": "45c1cc29130796f083425310aad7c10ed9c2a4cf7031e69dd40cd8551af65af2",
        "chunkIds": [
--- a/api/controllers/StorageController.js
+++ b/api/controllers/StorageController.js
@@ -14,7 +14,7 @@ class StorageController {
     var allFiles = [...new Set([...allUploadedFiles, ...allDownloadedFiles])];
     // return a subset of the attributes to the client
     const files = allFiles.map((file) =>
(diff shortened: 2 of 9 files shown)


jensendarren (approved):
This appears to work!

tankakatan (approved):

PR #127, opened 2022-01-17
Removing unimplementd exceptions

diogosmendonca:


jensendarren:
@diogosmendonca did you manage to test the storage API endpoints using Postman? What is needed in this story is to fetch the data from the point database and return via the API. The `throw` statements were placed there because the implementation is using Levvel DB and now we are using Sequelize. I can demo this today.

jensendarren (approved):
@diogosmendonca OK I just tested this in Postman and the APIs work! That was easier than I had thought :)

PR #128, opened 2022-01-17
Making --contract optional when using point-deploy from deployspace

diogosmendonca:


tankakatan:
Thanks @diogosmendonca, could you please retarget all your PR to `develop` instead.

jensendarren:
@diogosmendonca for this story we need to modify `deploy-from-deployspace.sh` file, not `deploy-sites.sh`. I'll revert this change and discuss with you.

jensendarren (changes requested):
The `point-deploy` script is no longer able to deploy the Smart Contracts when I tested this. I've updated the story with a 'definition of done' as follows (more details are on the story card): 

**Definition of Done**

* If executed like so: `point-deploy --contracts` then the contracts are deployed
* If executed like so `point-deploy` then the contracts are not deployed.

diogosmendonca:
I added the --contracts param to the point-deploy script. Now should works.

jensendarren (approved):

PR #136, opened 2022-01-19
update Identity contract solidity version to ^0.8.0

jensendarren:


jensendarren:
@tankakatan - I've made this change since there is a compile error in the `contract_deployer` container as follows:

`contract_deployer_1  | ParserError: Source file requires different compiler version (current compiler is 0.8.6+commit.11564f7e.Linux.g++) - note that nightly builds are considered to be strictly less than the released version
contract_deployer_1  |  --> project:/contracts/Identity.sol:2:1:
contract_deployer_1  |   |
contract_deployer_1  | 2 | pragma solidity 0.8.0;
contract_deployer_1  |   | ^^^^^^^^^^^^^^^^^^^^^^
`

tankakatan (approved):

PR #143, opened 2022-01-21
Fix/point social UI/UX

JatinBumbra:
PR for the [ticket](https://trello.com/c/t8Y0jpEF/9-in-pointsocialz-only-show-hand-cursor-on-post-in-clickable-areas)

Basic Gist: The pointer cursor should not cover the entire post section in the feeds, but only the intended elements - the poster's avatar, like button and comments link.


jensendarren (changes requested):
I notice in the CSS file the `post` class cursor has been commented out but I am not sure if that would satisfy the definition of done for this story? That is so that in the UI the cursor changes to a hand only on clickable areas. So basically in a post block the clickable areas are: 

- Avatar of the user who posted the update
- The 'Like' Icon
- The 'Comments' link

So when the users cursor is hovering over any of the above it should change to a hand to indicate that it is clickable.

JatinBumbra:
This fixes the hand-cursor problem over entire post. The individual elements already had the `cursor: pointer` which brings a hand-cursor when we hover. The actual problem was that the entire `post` class had a `cursor: pointer` so the entire post should a hand-cursor when hovered. So commenting this fixes it, alternatively we can remove this as well.

jensendarren:
Awesome thanks @JatinBumbra - I have it working now. I will merge into `develop`.

PR #150, opened 2022-01-24
Warinig when a zapp has missing dependencies

diogosmendonca:
Adding a warning when a Zapp has a package.json and not the folder node_modules. In this case the zapp developer should run npm install to install the dependencies before deploy the site.


jensendarren (approved):
Looks good! I also moved warning message up and added some color to make it stand out. I'll merge this into develop.

PR #151, opened 2022-01-24
adding the build script to the alias

diogosmendonca:
Adding a build script for e2e environment that consider the name of the branch as the tag of container image.


jensendarren (changes requested):
There is another step to run (step 4 in the story) which is have the script updates the local `.env.e2e` file by replacing `POINT_NODE_VERSION=` with the current latest Docker Tag (for  example `POINT_NODE_VERSION=develop`. The reason for this is so that once this image is built the developer can immediately run `point-e2e up -d` and it will use the tagged image that was just built. Can discuss further today along with your proposal of using the `--no-cache` option.

diogosmendonca:
> 
The step 4 is already implemented in [this line](https://github.com/pointnetwork/pointnetwork/pull/151/commits/5795c32296fecd3e9bc096dcee6096be07202e1c#diff-a8d2718217f616f636025de1040e98e98d3d3575f38f40899ea4988b015b22b3R11). Should it run even though the image is already built?

jensendarren (approved):

PR #159, opened 2022-01-27
Feature/migrate genesis data cli

AlexandreCarvalhoDeMelo:


scripts/blockchain/migrate-alpha-data.js
```diff
@@ -0,0 +1,140 @@
...
+]
+
+if(process.argv[2] === undefined){
+    console.log(`Please inform the zapp to migrate eg: blog.z`);
+    return;
+}
+
+const site = process.argv[2];
+
+init = () => {
```
tankakatan, line 71: Please don't declare variables in the global scope. Use `const` statements instead.

tankakatan (commented):
Please resolved the below issue. Also we will improve this scripts later on.

scripts/blockchain/migrate-alpha-data.js
```diff
@@ -0,0 +1,140 @@
...
+           }
+        ],
+        "name":"addMigrator",
+        "outputs":[],
+        "stateMutability":"nonpayable",
+        "type":"function"
+     }
+]
+
+init = () => {
```
tankakatan, line 64: I think I wasn't clear in my previous comment. You moved the `if` statement and some other code inside the `init` function, but that's now what I mean.

What I mean is that you declare the `init` and other functions in this file without `const` keyword, which adds these function to the app's global scope. This is an anti-pattern, and we should avoid it. Please add `const` keywords for each function declaration here.

tankakatan (changes requested):

jensendarren (approved):

brianunlam:
Hi team, now we are running an eslint check that blocks the merging if there are some errors. Locally run "npm run fix" and fix not auto-fixable errors manually and then update the pr until check passes. Also you can add eslint comments in order to avoid some particular errors. Any doubt don't hesitate to contact me.

PR #163, opened 2022-01-28
Refactor storage

Koochr:
Note: I didn't remove kademlia since it's used by other files, not ones I've been working with. So I'm afraid it can break something, and the changes are already quite large.


client/storage/index-new.js
```diff
@@ -0,0 +1,436 @@
...
+    const directory = await getJSON(dirId);
+    const segments = filePath.split(/[/\\]/).filter(s => s !== "");
+    const nextFileOrDir = directory.files.find(f => f.name === segments[0]);
+    if (!nextFileOrDir) {
+        throw new Error(`Failed to find file ${filePath} in directory ${dirId}: not found`);
+    }
+    if (segments.length === 1) {
+        return nextFileOrDir.id;
+    }
+    if (nextFileOrDir.type === "fileptr") {
```
tankakatan, line 421: Let's add type enum in order to avoid typos.

client/zweb/deployer/index.js
```diff
@@ -3,11 +3,16 @@ const fs = require('fs');
...
 
+// TODO: direct import cause fails in some docker scripts
+let storage;
+
 class Deployer {
     constructor(ctx) {
         this.ctx = ctx;
         this.config = this.ctx.config.deployer;
         this.cache_uploaded = {}; // todo: unused? either remove or use
+        const _storage = require("../../../client/storage/index-new.js");
```
tankakatan, line 14: As long as you have removed both `index.js` and `index-arweave.js` let's rename the file to just `index.js`.

client/zweb/deployer/index.js
```diff
@@ -3,11 +3,16 @@ const fs = require('fs');
...
+        storage = _storage;
```
tankakatan, line 15: Why do we need this intermediary variable `_storage`?
And also, why do you `require` the file in constructor. Is there a circular dependency?

db/models/chunk.js
```diff
@@ -1,242 +1,35 @@
...
 Chunk.init({
     id: { type: Sequelize.DataTypes.STRING, unique: true, primaryKey: true },
     size: { type: Sequelize.DataTypes.INTEGER, allowNull: true },
+    dl_status: { type: Sequelize.DataTypes.STRING, defaultValue: DOWNLOAD_STATUS.NOT_STARTED },
 
-    ul_status: { type: Sequelize.DataTypes.STRING, defaultValue: Chunk.UPLOADING_STATUS_CREATED },
-    dl_status: { type: Sequelize.DataTypes.STRING, defaultValue: Chunk.DOWNLOADING_STATUS_CREATED },
-
-    // allowNull in the next rows because this is only related to uploading and storing, not chunks queued for downloading
+    // TODO: not used, remove
```
tankakatan, line 23: I think we should use the upload status for chunks as well. The statuses I can think of right away are as follows:
* `in_progress` - needed to avoid attempts to download the same chunk twice
* `failed` - the upload may fail, and the retry logic may be added outside of the storage itself
* `uploaded` - indicates successful upload.

What do you think?

db/models/index.js
```diff
@@ -19,9 +19,10 @@ class SequelizeFactory {
             define: this.config.define,
             storage: this.config.storage,
             transactionType: this.config.transactionType,
-            retry: {
-                max: this.config.retry.max
-            },
+            // TODO: docker e2e fails bc this is missing in config. is it needed?
```
tankakatan, line 22: Please remove the comment and make sure the app still works.

docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: '3.9'
 
 services:
   point_node:
-    image: pointnetwork/pointnetwork_node:$POINT_NODE_VERSION
+    image: pointnode-local
```
tankakatan, line 5: Please avoid committing such things, otherwise it may brake the app completely for everyone.

tankakatan (changes requested):
Everything seems fine except, I am not sure we should get rid of upload/download statuses for both chunk and files. Let's discuss. Also please review my comments and fix what is possible to fix right away.

api/controllers/StorageController.js
```diff
@@ -18,45 +16,37 @@ class StorageController extends PointSDKController {
...
     async putString() {
         let data = this.payload.data;
         if(data) {
-            const cache_dir = path.join(this.ctx.datadir, this.config.cache_path);
-            utils.makeSurePathExists(cache_dir);
-            let tmpPostDataFilePath = path.join(cache_dir, utils.hashFnUtf8Hex(data));
-            fs.writeFileSync(tmpPostDataFilePath, data);
-            let uploaded = await this.ctx.client.storage.putFile(tmpPostDataFilePath);
-            return this._response(uploaded.id);
+            const id = await uploadFile(data);
```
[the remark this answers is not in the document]
tankakatan: I think this is another reason why we need to keep the statuses. For large files we'll have to be able to monitor the upload/download statuses, right?

client/zweb/deployer/index.js
Koochr: This is done for easier reviewing, otherwise you would see the diff between the old and new files. I am planning to rename it as a next step

client/zweb/deployer/index.js
Koochr: I think there was an issue with require, but I'll re-check

db/models/index.js
Koochr: It was not in e2e. In a normal or dev it worked fine, but I will also re-check

docker-compose.yaml
Koochr: oops, that was occasionally, I thought I removed it

api/controllers/StorageController.js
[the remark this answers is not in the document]
Koochr: As we discussed during the call, leaving this as is for now.

api/controllers/StorageController.js
```diff
@@ -68,26 +58,6 @@ class StorageController extends PointSDKController {
         return this._response(file);
     }
 
-    // Returns all chunk metadata stored in the nodes leveldb
-    async chunks() {
```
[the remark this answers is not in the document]
Koochr: As we discussed during the call, this method is of no use now, so removing it. I will re-implement it during the next steps, if needed.

client/storage/index-new.js
```diff
@@ -0,0 +1,436 @@
...
+    await chunk.save();
+
+    try {
+        const query = getDownloadQuery(
+            chunkId,
+            config.arweave_experiment_version_major,
+            config.arweave_experiment_version_minor
+        );
+
+        const queryResult = await request('https://arweave.net/graphql', query);
```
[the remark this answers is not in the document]
Koochr: There is no such URL anywhere in the config. I've put TODO, my suggestion is not to touch this unless we get to the config refactor.

client/storage/index-new.js
```diff
@@ -0,0 +1,436 @@
...
+            chunk.dl_status = DOWNLOAD_STATUS.COMPLETED;
+            await chunk.save();
+            return buf;
+        }
+
+        throw new Error("No mathing hash found in arweave");
+    } catch (e) {
+        logger.error(`Chunk download failed, chunk id: ${chunk.id}`);
+        logger.error(e);
+        chunk.dl_status = DOWNLOAD_STATUS.FAILED;
```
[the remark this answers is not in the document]
Koochr: Added a comment

client/storage/index-new.js
```diff
@@ -0,0 +1,436 @@
...
+        const {
+            type,
+            hash,
+            chunks,
+            filesize,
+            merkle: merkleHash
+        } = JSON.parse(chunkInfo.slice(CHUNKINFO_PROLOGUE.length + 1));
+
+        if (type !== 'file') {
+            throw new Error('Bad file type');
```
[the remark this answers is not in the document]
Koochr: As we discussed, to be implemented, I've added TODO

db/models/chunk.js
Koochr: As we discussed in the call, using single status for upload and download.

api/controllers/StorageController.js
```diff
@@ -18,45 +16,37 @@ class StorageController extends PointSDKController {
...
-                id,
-                originalPath,
-                size,
-                redundancy,
-                expires,
-                autorenew,
-                chunkCount: chunkIds.length,
-                ul_status
-            }))(file));
+        const files = await Promise.all(allFiles.map((file) =>
```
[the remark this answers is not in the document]
Koochr: This was a part of the old code, fixed it

client/storage/index-new.js
Koochr: Done

client/zweb/deployer/index.js
Koochr: Yep, if I move require to the top, e2e fails. You can see that other relative requires are also done dynamically.

db/models/index.js
Koochr: Also nothing changed. With this config docker e2e fails. Leaving it as is for now.

client/storage/index-new.js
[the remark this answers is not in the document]
tankakatan: > There is no such URL anywhere in the config. I've put TODO, my suggestion is not to touch this unless we get to the config refactor.

I mean, please add this url to the `defaultConfg.json` file next to `arweave_airdrop_endpoint` and get it from the config here instead of using the hardcoded string. Otherwise it cannot be changed when running the app in another environment (which I am already working on).

client/zweb/deployer/index.js
tankakatan: Noted on circular dependency. But why do we use the `_storage` variable then? Why not to write `storage = require("../../../client/storage/index-new.js");`?

Koochr: But I've already updated it

tankakatan (approved):

PR #168, opened 2022-01-31
Removing read only from volumes of deployment folders

diogosmendonca:
When a Zapp is deployed, point_node container downloads the files from arweave to run it locally. However, in website_owner container as the examples folder is mounted as ready only in e2e config, arweave cannot write to this folder. The same could be happening in YNet deployspace folder in the machine that deployed the zapp. 

This is the simpler solution that I found to solve this problem. However, we need to consider the potential side effects of setting the volumes as writable. One other problem with read only container for examples or deployspace folder is to run npm install. The node_modules folder is system dependent and in e2e config we are only able to run npm install from outside of the container, what possibly may lead to system incompatibilities in the dependencies.


jensendarren:
Not relevant anymore.

PR #180, opened 2022-02-03
Fix build-e2e script

JatinBumbra:
Fixes the issue with tagging when we have `/` in our branch. Great StackOverflow resource [here](https://stackoverflow.com/questions/62905914/turning-a-git-branch-name-into-a-valid-docker-image-tag) for reference.

### DIFF
--- a/.bash_alias
+++ b/.bash_alias
@@ -7,11 +7,14 @@ alias point-browser-docker-2="web-ext run --firefox-profile=website_owner_docker
 alias point-browser-docker-3="web-ext run --firefox-profile=website_visitor_docker --keep-profile-changes --source-dir dist/prod --url https://point/"
 
 build-e2e() {
-    docker image inspect pointnetwork/pointnetwork_node:`git branch --show-current` > /dev/null 2>&1;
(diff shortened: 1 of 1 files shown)


.bash_alias
```diff
@@ -7,11 +7,14 @@ alias point-browser-docker-2="web-ext run --firefox-profile=website_owner_docker
 alias point-browser-docker-3="web-ext run --firefox-profile=website_visitor_docker --keep-profile-changes --source-dir dist/prod --url https://point/"
 
 build-e2e() {
-    docker image inspect pointnetwork/pointnetwork_node:`git branch --show-current` > /dev/null 2>&1;
+    local image=$(git branch --show-current | sed 's/[^a-zA-Z0-9]/-/g');
```
jensendarren, line 10: @JatinBumbra - unfortunately `sed` is not working as expected on mac osx which is why we switched to using perl scripts (see this commit for details: 22a50528b176573f19dcc4d361090fba203dc451).

jensendarren (changes requested):

jensendarren (approved):

PR #191, opened 2022-02-08
Refactor wallet

Koochr:


client/zweb/renderer/index.js
```diff
@@ -82,7 +81,7 @@ class Renderer {
             },
             decrypt_data: async function (encryptedData, unparsedEncryptedSymmetricObjJSON) {
                 const host = this.host;
-                const {privateKey} = this.renderer.ctx.wallet.config;
+                const privateKey = this.renderer.ctx.wallet.getNetworkAccountPrivateKey();
```
brianunlam, line 84: I'm not sure, but I think the wallet should not provide the private key, just services signing with it.

tankakatan: @brianunlam, you're right, but we cannot avoid it completely ATM. In some cases we still need it. Not for decryption in particular, but there are two more places where we still need to get the private key. We'll think how to improve it soon. I already have some ideas.

point
```diff
@@ -86,7 +86,7 @@ if (program.verbose) ctx.config.log.level = 'trace';
 
 // ------------------- Init Logger ----------------- //
 
-ctx.log = logger.child({account: ctx.config.client.wallet.account});
+ctx.log = logger.child({module: "main"}); // TODO: account
```
tankakatan, line 89: 1. Why didn't `eslint` show the double quotes error here?
2. Let's rename this module to `point` instead of `main`

tankakatan (approved):

tankakatan (approved):

PR #200, opened 2022-02-14
Feature/refactor config

Koochr:


client/storage/index.js
```diff
@@ -14,14 +14,18 @@ const {promises: fs} = require('fs');
...
-const getDownloadQuery = (chunkId, majorVersion, minorVersion) =>
+// TODO: possibly split this file into several ones after migrating to modules
+const getDownloadQuery = chunkId =>
     gql`{
       transactions(
         tags: [
           {
-            name: "__pn_chunk_${majorVersion}.${minorVersion}_id",
+            name: "__pn_chunk_${config.get(
+        'storage.arweave_experiment_version_major'
```
brianunlam, line 27: can we add these versions in a variable/const in order not to have to get them from config everytime they are needed? They are used below many times.

client/storage/index.js
```diff
@@ -103,13 +91,9 @@ const getChunk = async (chunkId, encoding = 'utf8', useCache = true) => {
...
 
-        const query = getDownloadQuery(
-            chunkId,
-            config.arweave_experiment_version_major,
-            config.arweave_experiment_version_minor
-        );
+        const query = getDownloadQuery(chunkId);
 
-        const queryResult = await request(config.arweave_gateway_url, query);
+        const queryResult = await request(config.get('storage.arweave_gateway_url'), query);
```
brianunlam, line 96: same for url, it's used after

client/storage/index.js
```diff
@@ -210,7 +202,7 @@ const uploadChunk = async data => {
 const uploadFile = async data => {
     const buf = Buffer.isBuffer(data) ? data : Buffer.from(data);
 
-    const CHUNK_SIZE = config.chunk_size_bytes;
+    const CHUNK_SIZE = config.get('storage.chunk_size_bytes');
```
brianunlam, line 205: same

network/web3bridge.js
```diff
@@ -40,11 +41,9 @@ class Web3Bridge {
     constructor(ctx) {
         this.ctx = ctx;
         this.log = ctx.log.child({module: 'Web3Bridge'});
-        this.connectionString = this.ctx.config.network.web3;
-        this.network_id = this.ctx.config.network.web3_network_id;
-        this.chain_id = this.ctx.config.network.web3_chain_id;
+        this.connectionString = config.get('network.web3');
         this.address = this.ctx.wallet.getNetworkAccount();
-        this.web3_call_retry_limit = this.ctx.config.network.web3_call_retry_limit || 4;
+        this.web3_call_retry_limit = config.get('network.web3_call_retry_limit') || 4;
```
brianunlam, line 46: can we use a constant here?

```
// top of file
const DEFAULT_WEB3_RETRY_LIMIT =4 

// ...
 
this.web3_call_retry_limit = config.get('network.web3_call_retry_limit') || DEFAULT_WEB3_RETRY_LIMIT;
```

point
```diff
@@ -87,17 +81,38 @@ program
...
-// todo: mix other param arguments into config, like here:
-// let argv;
-// let config = rc('kadence', options(program.datadir), argv);
+const ctx = {};
 
-if (!program.config) {
-    program.config = path.join(program.datadir, 'config.json');
+if (program.datadir) {
+    process.env.DATADIR = program.datadir;
 }
```
brianunlam, line 88: not sure what happens here if program.datadir is not defined.

Also we can use:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment

point
```diff
@@ -87,17 +81,38 @@ program
...
+
+    for (const network in networks) {
+        const {address} = networks[network];
+        if (address && typeof address === 'string') {
+            return address;
+        }
+    }
+}
+
+if (process.env.MODE === 'e2e') {
```
brianunlam, line 107: I think for MODE we should have a file with all the available MODES in a constant

process.env.MODE === E2E

or

process.env.MODE === mode.E2E

point
```diff
@@ -175,7 +182,8 @@ if (program.makemigration) {
 if (program.migrate) {
     const seq_cmd = program.migrate_undo ? 'db:migrate:undo' : 'db:migrate';
     exec(
-        `npx sequelize-cli ${seq_cmd} --config resources/sequelizeConfig.json --env ${process.env.DB_ENV || ctx.config.db.database}`,
+        `npx sequelize-cli ${seq_cmd} --url sqlite:${
+            path.join(config.get('datadir'), config.get('db.storage'))} --env ${config.get('db.env')}`,
```
brianunlam, line 186: datadir is used after

brianunlam (commented):
LGTM

point
```diff
@@ -189,14 +197,6 @@ if (program.migrate) {
     return;
 }
 
-// ------------------ Remove Everything ------------ //
-if (program.debug_destroy_everything) {
```
tankakatan, line 193: This option is something what could be used by some other devs, especially when running the app outside of docker. It's not right to get rid of it without letting anyone know.

wallet/index.js
```diff
@@ -73,42 +75,45 @@ class Wallet {
...
         this.transactionEventEmitter = new events.EventEmitter();
     }
 
     async start() {
-        this.keystore_path = path.join(this.ctx.datadir, this.config.keystore_path);
-        if (!fs.existsSync(this.keystore_path)) {
-            fs.mkdirSync(this.keystore_path, {recursive: true});
+        await makeSurePathExistsAsync(this.keystorePath);
+        try {
+            this.#secretPhrase = JSON.parse(
```
tankakatan, line 87: It'd be much simpler to write `this.#secretPhrase = require(path.join(this.keystorePath, 'key.json')).phrase;` instead.

wallet/index.js
```diff
@@ -73,42 +75,45 @@ class Wallet {
...
+                await fs.promises.readFile(path.join(this.keystorePath, 'key.json'), 'utf-8')
+            ).phrase;
+        } catch (e) {
+            if (e.code === 'ENOENT') {
+                this.#secretPhrase = '';
```
tankakatan, line 92: Why should we handle this case? Running the app without the phrase makes no sense.

point
```diff
@@ -87,17 +81,38 @@ program
...
+        if (address && typeof address === 'string') {
+            return address;
+        }
+    }
+}
+
+if (process.env.MODE === 'e2e') {
+    process.env.IDENTITY_CONTRACT_ADDRESS = getContractAddress('Identity');
+    process.env.STORAGE_PROVIDER_REGISTRY_CONTRACT_ADDRESS = getContractAddress('StorageProviderRegistry');
+}
```
tankakatan, line 110: I think you only tested the app in the `dev` mode where you have the `truffle/` dir mounted. I bet it won't work in the `production` mode. From what I can see, you forgot to compile the `Identity` contract and store its' ABI in the `/truffle/build/contracts` dir. The app will not work without it. Please see the `compilePointContracts` in the deleted `patch-config.js` file. The function should run in all environments except the `e2e` one.

Please restore the function and add it along with the `getContractAddress` one to the `web3bridge` or a separate module importing it here.

package.json
```diff
@@ -75,19 +77,16 @@
...
     "test:db:docker": "printf '\\e[2J\\e[3J\\e[H' && npm run migrate && NODE_ENV=development npm run test",
     "test:restart": "docker-compose -f docker-compose.test.yaml restart tests && docker logs -f pointnetwork_tests",
-    "start": "./point --datadir $DATADIR -v",
-    "start:dev": "nodemon ./point -- --datadir $DATADIR -v",
-    "start:docker": "./scripts/patch-config.js && npm run migrate && npm start",
-    "start:docker:dev": "./scripts/patch-config.js && npm run migrate && npm run start:dev",
-    "start:docker:e2e": "./scripts/patch-config-e2e.js && npm run migrate && npm run start:dev",
-    "migrate": "./point migrate --datadir $DATADIR",
+    "start": "./point",
+    "start:docker": "./point migrate && ./point",
```
tankakatan, line 81: Let's reuse the scripts here and one line below. Like `"start:docker": "npm run migrate && npm start"`.

network/web3bridge.js
tankakatan: We no longer need the default value here. It is already specified in the `config/default.yaml` file.

docker-compose.yaml
```diff
@@ -4,32 +4,15 @@ services:
   point_node:
     image: pointnetwork/pointnetwork_node:$POINT_NODE_VERSION
     container_name: point_node
-    entrypoint: 'npm'
-    command: 'run start:docker'
     volumes:
       - point_contracts:/app/truffle/build/contracts
       - point_node_data:/data
       - $POINT_KEYSTORE:/data/keystore
-      - $POINT_DEPLOYSPACE:/app/deployspace:ro
```
tankakatan, line 13: Why did you decide to remove this?

point
tankakatan: Nullish assignment is not applicable here, and the app will work fine if the `program.datadir` is not defined, since the default value from the `config/*.yaml` will be used.

tankakatan (changes requested):

point
Koochr: This function is not needed since the contracts are compiled by `deployer` in e2e mode. And you mentioned that it's only used in development?
I also tried running app in production and it works. Or did I get something wrong?

tankakatan: > This function is not needed since the contracts are compiled by `deployer` in e2e mode. And you mentioned that it's only used in development?
The contract is compiled by the deployer only in the `e2e` mode. In all other modes we need to compile it or get it from a remote location which is not implemented at this point.
> I also tried running app in production and it works. Or did I get something wrong?
Ok then this could be because you've built an image with the `truffle/build/contracts` folder copied into container, which is wrong. Please add this folder to `.dockerignore`. Also mind it is already ignored by git, and that's for a purpose.

docker-compose.yaml
tankakatan: > It's never used anywhere.

From what I can see this is not true. Please take a look at the `docker-compose.yaml` and `deploy-from-deployspace.sh` files.

wallet/index.js
Koochr: App is often run without a phraze and it was quite a long time ago, I was checking it. It was never created in docker before starting the app

point
```diff
@@ -197,7 +205,74 @@ if (program.migrate) {
...
+
+    log.info('Compiling contracts...');
+
+    const buildPath = path.resolve(__dirname, 'truffle', 'build', 'contracts');
+    if (!existsSync(buildPath)) {
+        mkdirSync(buildPath);
+    }
+    const contractsPath = path.resolve(__dirname, 'truffle', 'contracts');
+
+    Promise.all(CONTRACT_ADDRESSES.map(async name => {
```
tankakatan, line 240: This solution is based on a slightly outdated version of the `compileContracts` function. Please see the most recent one [here](https://github.com/pointnetwork/pointnetwork/blob/develop/scripts/patch-config.js#L25). The main difference is that we don't want to re-compile the contracts if they are already compiled. The corresponding test takes place [here](https://github.com/pointnetwork/pointnetwork/blob/develop/scripts/patch-config.js#L41). Please double check for any other updates to the function code.

tankakatan (approved):

PR #213, opened 2022-02-18
Start ts project, add src folder

Koochr:
Important note: you should change path to `pac.js` file in firefox settings to make proxy work


.eslintrc.js
```diff
@@ -4,29 +4,27 @@ module.exports = {
...
+        '@typescript-eslint',
         // 'import',
         'filenames',
         'prefer-arrow'
     ],
     ignorePatterns: [
         'node_modules',
         '*.min.js',
-        'example/*',
+        'src/example/*',
```
tankakatan, line 25: Let's not move examples directory to `src`. All the examples are completely different projects which may be also written in `ts` (and some of them actually are) implying different `ts` configuration.

package.json
```diff
@@ -59,32 +61,37 @@
...
-    "start:docker": "npm run compile && npm run migrate && npm start",
-    "start:docker:e2e": "npm run migrate && MODE=e2e npm start",
+    "start": "node dist/index.js",
+    "start:docker": "npm run compile && npm run migrate && npm run build && npm start",
+    "start:docker:dev": "npm run compile && npm run migrate && npm run start:dev",
+    "start:docker:e2e": "npm run migrate && npm run build && MODE=e2e npm start",
     "start:dev": "nodemon",
-    "migrate": "./point migrate",
-    "compile": "./point compile",
+    "build": "tsc && cp -R src/truffle dist/truffle && cp -R src/internal dist/internal && cp -R src/client/proxy/certs dist/client/proxy/certs && cp src/client/proxy/ca.srl dist/client/proxy/ca.srl",
```
tankakatan, line 92: We won't need to copy these dirs to `dist` if we don't store them in src as suggested in the above comment. Also certs should probably be placed to resources I think.

cli
```diff
@@ -1,39 +1,17 @@
 #!/usr/bin/env node
```
tankakatan, line 1: Why did you decide to rename this file? I don't see how it is related to the task.

src/index.ts
```diff
@@ -0,0 +1,153 @@
...
+const program: any = new Command();
+
+// Disable https://nextjs.org/telemetry
+process.env['NEXT_TELEMETRY_DISABLED'] = '1';
+
+// TODO: Enabled this option for backward-compatibility support, but remove later to support newer syntax
+program.storeOptionsAsProperties();
+
+program.version(`
+    0.0 alpha
```
tankakatan, line 16: Let's pick the version from `package.json`.

src/threads/decrypt.js
```diff
@@ -3,7 +3,7 @@
 const fs = require('fs');
 const crypto = require('crypto');
 const utils = require('#utils');
-const defaultConfig = require('../resources/defaultConfig.json');
+const defaultConfig = require('../../resources/defaultConfig.json');
```
tankakatan, line 6: BTW why are we still using `defaultConfig.json` file instead of `config` module here and in the `encrypt.js` below?

src/truffle/.gitignore
```diff
@@ -0,0 +1,2 @@
+build/contracts
+contracts/Hello.sol
```
tankakatan, line 2: Instead of having `Hello.sol` here, let's come up with a more general purpose solution, like for example keeping your personal stuff in a `src/truffle/local` dir or something.

cli
```diff
@@ -84,37 +62,11 @@ if (program.datadir) {
...
-const logger = require('./core/log');
-const Model = require('./db/model');
-const Deployer = require('./client/zweb/deployer');
-const DB = require('./db');
-
-ctx.basepath = __dirname;
+const logger = require('./src/core/log');
+const Console = require('./src/console');
+const Deployer = require('./src/client/zweb/deployer');
+const DB = require('./src/db');
```
tankakatan, line 69: Looks like we don't need `logger`, `Deployer` and `DB` imports here. Also I am not sure the `Console` import will work when we rename the file to `ts` in the `src` directory.

tankakatan (changes requested):

cli
Koochr: I've split this file to the server and cli. This is done bc cli triggers actions such as migrate or compile, which should be done regardless of ts compilation. Otherwise we would have cli actions both in src and dist, both in ts and js, which is not convenient at all.

src/threads/decrypt.js
Koochr: This is not used anywhere, this replacement is done automatically. Should I remove them?

cli
Koochr: We do as we have logger below (well, we can replace it with console). Deployer and DB are needed for deploy and migrate scripts.
I didn't change this, the file was just split

.eslintrc.js
Koochr: scripts, resources, config and tests are not in src. This is probably a forgotten change.
Not sure about truffle and internal, it depends on how are they used. If we import some of these files anywhere, we can't move them out of src.

tankakatan (approved):

PR #221, opened 2022-02-21
Modify the way utils is required to match pkg requirements

brianunlam:
Vercel/Pkg doesn't support imports on package.json and require utils the way we are currently doing it. In order to address that we are requiring utils the old way.


tankakatan (approved):

package.json
```diff
@@ -112,7 +112,6 @@
     "lint:sol": "solhint 'example/**/contracts/*.sol' 'truffle/contracts/*.sol'"
   },
   "imports": {
```
tankakatan, line 114: Let's remove or clean up the entire `imports` section in order to avoid using or adding other aliases declared here unless we find a solution to this issue.

PR #222, opened 2022-02-21
Use custom path config folder when running in pkg mode

brianunlam:
When running inside the packaged version the configuration should be retrieved from the internal packaged config by default config library uses process.cwd() to reference the config folder.
When using vercel/pkg process.cwd references real folder and not packaged folder overwriting this env variable fixes the problems

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,6 +4,18 @@ import {existsSync, writeFileSync} from 'fs';
 import lockfile from 'proper-lockfile';
 import disclaimer from './disclaimer';
 
+export const RUNNING_PKG_MODE = (process as any).pkg ? true : false;;
+
+if (RUNNING_PKG_MODE) {
+  // when running inside the packaged version the configuration should be
+  // retrieved from the internal packaged config
+  // by default config library uses process.cwd() to reference the config folder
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -4,6 +4,18 @@ import {existsSync, writeFileSync} from 'fs';
...
 
+export const RUNNING_PKG_MODE = (process as any).pkg ? true : false;;
+
+if (RUNNING_PKG_MODE) {
+  // when running inside the packaged version the configuration should be
+  // retrieved from the internal packaged config
+  // by default config library uses process.cwd() to reference the config folder
+  // when using vercel/pkg process.cwd references real folder and not packaged folder
+  // overwriting this env variable fixes the problems
+  process.env.NODE_CONFIG_DIR = path.resolve(__dirname, 'config');
```
tankakatan, line 15: Wouldn't it be better to define this variable in the `npm start` script? Something like:
```
  "start": "NODE_CONFIG_DIR=./config node dist/index.js",
```
I am not sure it will work, so that's just a suggestion. Please let me know if you already tried it or if there's another problem I am missing here.

tankakatan (approved):

PR #226, opened 2022-02-21
Remove kadence and add path requirement to fix utils

brianunlam:


src/core/utils.js
```diff
@@ -137,8 +138,18 @@ const utils = {
         return results;
     },
 
-    urlToContact: function (str) {
-        return kadUtils.parseContactURL(str);
+    urlToContact: function (addr) {
```
brianunlam, line 141: references: 

https://github.com/pointnetwork/kadence/blob/b50cb32a5d0ff45ea4c799b232f053ab21c6f8f3/lib/utils.js#L314

src/core/utils.js
```diff
@@ -137,8 +138,18 @@ const utils = {
...
+        const {protocol, hostname, port, hash} = url.parse(addr);
+        const contact = [
+            (hash ? hash.substr(1) : null) ||
+            Buffer.alloc(constants.B / 8).fill(0).toString('hex'),
```
tankakatan, line 145: I think this function may fail as long as the `constants` variable is undefined.

But basically I propose to remove this `urlToContract` function completely. It is not used anywhere in the code and only makes sense in context of p2p communication within Kademlia network which we decided to give up on.

tankakatan (changes requested):

src/core/utils.js
brianunlam: nice

tankakatan (approved):

PR #229, opened 2022-02-21
start hardhat transition

elblockdude:


deployer.Dockerfile
```diff
@@ -4,12 +4,48 @@ FROM node:14.17.5-alpine
...
-web3@1.5.2
+RUN apk update && apk add --no-cache git
+
+#RUN npm i -g truffle && \
+#truffle init && \
+#npm i @openzeppelin/contracts@4.3.0 \
+#@truffle/hdwallet-provider \
+#web3@1.5.2
+
+RUN yarn add --dev "hardhat@^2.8.4" \
```
jensendarren, line 17: @elblockdude can these dependencies be installed via the hardhat/package.json file that is already defined?

docker-compose.e2e.yaml
```diff
@@ -26,25 +26,29 @@ services:
         --account 0x011967d88c6b79116bb879d4c2bc2c3caa23569edd85dfe0bc596846837bbc8e,0x56bc75e2d63100000
 
   contract_deployer:
-    image: pointnetwork/pointnetwork_deployer:dev
     container_name: deployer
+    build:
```
jensendarren, line 30: @tankakatan building the image from the compose service definition is ok going forward or should we deploy a ready made image to DockerHub?

hardhat/build/contracts/Identity-address.json
```diff
@@ -0,0 +1 @@
+{"address":"0x57BD87BD93b5377Cf012EC8adFb075E6514256E1"}
```
jensendarren, line 1: @elblockdude for which network?

deployer.Dockerfile
tankakatan: Definitely this should be replaced with `npm i`.

deployer.Dockerfile
```diff
@@ -4,12 +4,48 @@ FROM node:14.17.5-alpine
...
+ "@typescript-eslint/parser@^4.29.1" \
+ "@types/chai@^4.2.21" \
+ "@types/node@^12.0.0" \
+ "@types/mocha@^9.0.0" \
+ "ts-node@^10.1.0" \
+ "typechain@^5.1.2" \
+ "typescript@^4.5.2"
+
+
+ENTRYPOINT ["npx"]
```
tankakatan, line 50: Also the entry point should self sufficient. At least it should be set to `npm start` or similar.

docker-compose.e2e.yaml
tankakatan: We should deploy it. Building images in compose files is fine during development. But the e2e compose is used not only by Point contract devs, but also by other teams. It would be much more convenient for them to download the image rather than build it from scratch.

hardhat/.gitignore
```diff
@@ -0,0 +1,19 @@
+node_modules
```
tankakatan, line 1: * Please do not add build artefacts as well as deployed contract addresses to the repository. The `hardhat/build` path should be ignored.
* Also I propose to avoid using tech-specific names for directories. Instead of `truffle` or `hardhat` the folder should be called `blockchain` or `backend` or something like that.

src/index.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
tankakatan, line 2: Looks like this line was added by mistake.

src/network/web3bridge.js
```diff
@@ -62,18 +61,18 @@ class Web3Bridge {
...
 
-    async start() {}
+    async start() { }
 
     async loadPointContract(contractName, at) {
+
         if (!(contractName in abisByContractName)) {
+
             const buildDirPath = path.resolve(
-                resolveHome(config.get('datadir')),
```
tankakatan, line 70: The artefacts should be stored in the `datadir` for now. Later the whole workflow here will be changed.

tankakatan (changes requested):

hardhat/build/contracts/Identity-address.json
chefcosmo: Hey Darren this will be added at every deploy cz we don't have network section inside the ABI anymore

deployer.Dockerfile
chefcosmo: Well i'll see if that works and come back to you guys

hardhat/build/contracts/Identity-address.json
elblockdude: Goes with whatever network is setup on your hardhat config file

PR #231, opened 2022-02-22
removing arlocal workaround

diogosmendonca:
Arlocal presented a bug for download chunks, [this one](https://github.com/textury/arlocal/issues/63). We bypassed the bug with a workarount to fetch chunks in another way. Now arlocal bug is fixed and we were able to remove the workaround. To test this PR is necessary to remove the image textury/arlocal and download it again from docker hub.

### DIFF
--- a/docker-compose.zappdev.yaml
+++ b/docker-compose.zappdev.yaml
@@ -17,7 +17,7 @@ services:
           service: contract_deployer
     volumes:
       - shared_contracts_zappdev:/build:rw
-    
+
   point_node:
     extends:
       file: docker-compose.base.yaml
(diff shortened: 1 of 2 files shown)


docker-compose.zappdev.yaml
```diff
@@ -75,7 +75,7 @@ services:
       DATADIR: /data
 
   arlocal:
-    image: textury/arlocal
+    image: textury/arlocal:v1.1.21
```
jensendarren, line 78: @diogosmendonca Note - I decided to add the version declaration explicitly here.

jensendarren (approved):

PR #233, opened 2022-02-23
Add workflow to create a  new release and attached packaged application for linux, mac and win

brianunlam:
It will listen for tags with the format v* to create a new release.

### DIFF
--- a/.github/workflows/pkg.yaml
+++ b/.github/workflows/pkg.yaml
@@ -1,0 +1,91 @@
+# This workflow will package the application
+
+name: Create release
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+
+
+jobs:
+  build:
+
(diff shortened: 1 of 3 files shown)


tankakatan (approved):

.github/workflows/pkg.yaml
```diff
@@ -0,0 +1,88 @@
...
+        draft: false
+        prerelease: false
+
+    - name: upload linux artifact
+      uses: actions/upload-release-asset@v1
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      with:
+        upload_url: ${{ steps.create_release.outputs.upload_url }}
+        asset_path: ./bin/pointnetwork-linux.gz
```
tankakatan, line 68: Should we also include version in the asset/executable filenames?

tankakatan (approved):

.github/workflows/pkg.yaml
brianunlam: sure

PR #235, opened 2022-02-23
Create separate keystore module to manage keys

brianunlam:


src/core/log.js
```diff
@@ -18,7 +19,11 @@ if (sendLogs && sendLogsTo) {
 }
 
 streams.push(
-    {level: options.level, stream: pino({prettyPrint: {colorize: true}})[pino.symbols.streamSym]},
+    {
+        level: options.level,
+        stream: pino({prettyPrint: {colorize: true}})[pino.symbols.streamSym],
+        address: getNetworkAddress()
```
tankakatan, line 25: This doesn't add address to each log message like we need. Instead we should do something like that:
```
streams.push(...);

let logger = pino(options, multistream(streams));

try {
  const account = getNetworkAddress();
  logger = logger.child({account});
} catch (e) {
  // no account address
}

module.exports = Object.assign(logger, {
  ....
});
```

src/wallet/keystore.ts
```diff
@@ -0,0 +1,58 @@
...
+            const secretPhrase = require(path.join(keystorePath, 'key.json')).phrase;
+            const hdwallet = hdkey.fromMasterSeed(bip39.mnemonicToSeedSync(secretPhrase));
+            wallet = hdwallet.getWallet();
+        }
+        return wallet;
+    };
+}
+
+function getArweaveKeyFactory() {
+    let arweaveKey: string|undefined|null = null;
```
tankakatan, line 24: Let's not mix `undefined` and `null` in the same variable. The logic should be the same as in `getWalletFactory`. Also arweave key is not a string. It is an object. Please generate one via `scripts/keys.js` and see the key structure.

src/wallet/keystore.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    getNetworkPublicKey() {
+        getWallet().getPublicKey().toString('hex');
+    },
+
+    getNetworkPrivateKey() {
+        getWallet().getPrivateKey().toString('hex');
+    },
+
+    getStorageKey() {
```
tankakatan, line 54: Let's rename this method to `getStroragePrivateKey`. Also the `getStorageAddress` method is missing here. You can create one using `jwkToAddress` method available in the `arweave` package.

src/wallet/keystore.ts
```diff
@@ -0,0 +1,58 @@
...
+    };
+}
+
+function getArweaveKeyFactory() {
+    let arweaveKey: string|undefined|null = null;
+    return (): string|undefined => {
+        if (arweaveKey === null) {
+            try {
+                arweaveKey = require(path.join(keystorePath, 'arweave.json'));
+            }catch(e){
```
tankakatan, line 29: Could you please check the linter rules for the syntax like that? I was under impression that it should require spaces between the brackets here, so that the statement should look like
```
} catch (e) {
```

tankakatan (changes requested):
Please also move the methods exposed by the `keystore` module from the `wallet` and make  other modules use them directly from the `keystore` instead.

src/core/log.js
```diff
@@ -6,23 +6,37 @@ const {multistream} = require('pino-multi-stream');
...
 
 if (sendLogs && sendLogsTo) {
     const [address, port] = sendLogsTo.split('://').pop().split(':');
     streams.push(new udpTransport({address, port}));
 }
 
 streams.push(
-    {level: options.level, stream: pino({prettyPrint: {colorize: true}})[pino.symbols.streamSym]},
+    {
+        level: RUNNING_PKG_MODE ? DEFAULT_LOG_LEVE_FOR_PRODUCTION : options.level,
```
tankakatan, line 24: We don't need to do this ternary here, as well as hard code the log level using `DEFAULT_LOG_LEVE_FOR_PRODUCTION` variable in this file. We should leverage `config` instead. Just override the config level in [this file](https://github.com/pointnetwork/pointnetwork/blob/develop/config/production.yaml), the same way it is set in the default config [here](https://github.com/pointnetwork/pointnetwork/blob/develop/config/default.yaml#L20). Then the dashboard will simply set the mode to `production` and it will work as intended.

brianunlam: so basically we should add a production.yaml file?

tankakatan: It is already added, just let's specify the log level there explicitly.

PR #240, opened 2022-03-01
Implement storage upload queue, retry count and request timeout

Koochr:


package.json
```diff
@@ -13,6 +13,9 @@
     "@openzeppelin/contracts": "^4.4.2",
     "@solana/web3.js": "^1.29.2",
     "@truffle/hdwallet-provider": "^1.5.0",
+    "@types/config": "^0.0.41",
```
Koochr, line 16: IDK why this changes appeared, they should be in develop already as they were added with typescript migration

Koochr: I'm not sure. We have typescript in main dependencies, why should types go to dev dependencies?
Also, I know approach, when docker is not installing dev dependencies, but it's also not our case.

tankakatan: Please move typescript to `devDependencies`. `dependencies` should only contain packages that are used in the app's runtime.

src/db/models/index.ts
```diff
@@ -30,7 +30,7 @@ class SequelizeFactory implements SequelizeFactory {
                 transactionType: this.config.transactionType,
                 retry: {max: this.config.retry.max},
                 logQueryParameters: true,
-                logging: log.debug.bind(log)
+                logging: config.get('db.logging') ? log.debug.bind(log) : false
```
Koochr, line 33: I've added this, bc now db floods console with debug messages. Default behavior is not changed, but now you can override it by local config file

src/util/migrate.ts
```diff
@@ -27,20 +29,21 @@ export default (async () => {
...
         storage: new SequelizeStorage({sequelize}),
-        logger: logger.child({module: 'migrate'})
+        logger: log
     });
 
     try {
         await umzug.up();
         await sequelize.close();
     } catch (e) {
+        log.error(e);
```
Koochr, line 46: It was super ambiguous without this log when migration failed. We should notify user here.

src/db/models/index.ts
tankakatan: This makes sense, but please be more specific introducing new config options. This particular one only controls the database logging, right? Please rename it appropriately. And also set it to `false` in `production.yaml`.

tankakatan (approved):

PR #242, opened 2022-03-02
Fix/PN-62

tankakatan:
[Fix internal dir issue appearing in the point executable runtime](https://point-labs.atlassian.net/browse/PN-62)


brianunlam (commented):
LGTM

PR #243, opened 2022-03-02
Remove logstash and elk stack configuration 

brianunlam:
It adds a new configuration for elk stack enabling security and setting certificates.


tankakatan:
All good. Just let's move it [here](https://github.com/pointnetwork/point-logstash).

PR #246, opened 2022-03-07
Fix gateway in e2e

Koochr:


docker-compose.storage.yaml
```diff
@@ -21,11 +21,9 @@ services:
       - arweave_data:/mnt/arweave-data
 
   gateway:
-    image: lucaarweave/arweave-gateway:0.0.1
+    image: INSERT_IMAGE_NAME_HERE
```
Koochr, line 24: @tankakatan Please don't forget to change the image name once you build and upload the fixed storage to the repo

PR #256, opened 2022-03-09
Move required folders check to a single file

Koochr:


src/client/storage/index-arlocal.js
```diff
@@ -64,12 +63,10 @@ const filesDir = path.join(resolveHome(config.get('datadir')), config.get('stora
 
 let arweave;
 // load the arweave key for arlocal
-const keystorePath = path.join(resolveHome(config.get('datadir')), config.get('wallet.keystore_path'));
+const keystorePath = config.get('wallet.keystore_path');
```
tankakatan, line 66: This should also needs `resolveHome`, since the path may be inside the home dir `~`. Home dir doesn't work when running the executable, that's why we need `resolveHome` call.

src/initFolders.ts
```diff
@@ -0,0 +1,35 @@
+import {makeSurePathExists} from './core/utils';
```
tankakatan, line 1: Please put this file to `src/util`. We'll move all the utility function there soon.

src/initFolders.ts
```diff
@@ -0,0 +1,35 @@
...
+import path from 'path';
+import config from 'config';
+
+const initFolders = async () => {
+    const datadir: string = config.get('datadir');
```
tankakatan, line 6: Please use `resolveHome` here. And also please double-check all the other places.

src/initFolders.ts
```diff
@@ -0,0 +1,35 @@
...
+    try {
+        await makeSurePathExists(datadir);
+    } catch (e) {
+        throw new Error('Datadir folder does not exist. Did you create it? Please, refer to this guide: https://github.com/pointnetwork/pointnetwork-dashboard/blob/main/ALPHA.md');
+    }
+
+    try {
+        await makeSurePathExists(config.get('wallet.keystore_path'));
```
tankakatan, line 14: `resolveHome`

src/initFolders.ts
```diff
@@ -0,0 +1,35 @@
...
+    } catch (e) {
+        throw new Error('Keystore folder does not exist. Did you create it? Please, refer to this guide: https://github.com/pointnetwork/pointnetwork-dashboard/blob/main/ALPHA.md');
+    }
+
+    try {
+        await makeSurePathExists(path.join(config.get('wallet.keystore_path'), 'key.json'));
```
tankakatan, line 20: Need to `resolveHome` every time we reference user-specified paths.

src/wallet/keystore.ts
```diff
@@ -8,15 +7,13 @@ import {JWKInterface} from 'arweave/node/lib/wallet';
 
 const arweave = Arweave.init({});
 
-const keystorePath = path.join(resolveHome(config.get('datadir')), config.get('wallet.keystore_path'));
+const keystorePath: string = config.get('wallet.keystore_path');
```
tankakatan, line 10: `resolveHome`

tankakatan (changes requested):

tankakatan (approved):

PR #264, opened 2022-03-11
PN-75 Improve Blockchain module (previously web3bridge).

germanviescas:
The goal was to have the `web3brige` module (now renamed to `blockchain`) be the only one to directly interact with the `web3` instance. Other modules, such as `wallet` no longer access web3 directly as they used to (through `this.ctx.web3` for example).

Also, `blockchain` was turned into an object instead of a class and web3 is only accesible by the "blockchain object" returned by the `createBlockchain` factory function. `blockchain` is exposed via `this.ctx.blockchain`.

To test the changes, I run the `e2e` environment several times with the `docker-compose.e2e.yaml` config, and connected to it with a Firefox profile using _localhost:8666_ as proxy.

I deployed 3 zapps: twitter, blog and pointsocial.

**Twitter**:

- Everything looks fine. I was able to post tweets and to like them.
twitter

**Blog**:
- Everything looks fine. I was able to create a blog post and write a couple of comments.
blog

**Point Social**:
- Doesn't look good. But I get the exact same thing when running the environment from the `develop` branch with latest changes. So, either I'm deploying it incorrectly, or there's an issue in `develop` already.
- These are the steps I followed:
  - cd example/pointsocial.z/
  - npm ci
  - npm run build
  - cd ../..
  - docker compose -f docker-compose.e2e.yaml up -d
  - docker exec -it point_node bash
  - ./point deploy example/pointsocial.z --contracts 
 pointsocial
 
 None of the steps above through any errors that I could see. The contract deployment seems to finish successfully. But I did find some timeout errors in the docker logs.
Screenshot from 2022-03-11 18-20-42

(The screenshot above is from `develop`, but I see the exact same thing running this branch).


src/network/blockchain.js
```diff
@@ -42,81 +43,61 @@ function createWeb3Instance({blockchainUrl, privateKey}) {
...
                     fs.mkdirSync(buildDirPath, {recursive: true});
                 }
-                const contractPath = path.resolve(
-                    this.ctx.basepath,
-                    '..',
-                    'truffle',
-                    'contracts'
-                );
-                await compileContract({name: contractName, contractPath, buildDirPath});
+                const contractPath = path.resolve(basepath, '..', 'truffle', 'contracts');
```
tankakatan, line 66: This is the only part where we use the `basepath`, but because of it we have to wrap the entire module into the `createBlockchain` function.

Let's make the `basepath` an argument to the `loadPointContract` function instead. It should also have a default value. Let's also compute it here so that it would be the same as the one we pass to the `createBlockchain` function.

The goal is to get rid of `createBlockchain` function completely. Instead we should simply export all the `blockchain` object methods, and each method should use a locally created `web3` closure.

src/network/index.js
```diff
@@ -32,8 +32,8 @@ class Network {
         // todo: just use rejectUnauthorized: false on transport-https, extend it
         // process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
 
-        this.web3bridge = new Web3Bridge(this.ctx);
-        await this.web3bridge.start(); // todo: do we need await?
+        this.blockchain = createBlockchain(this.ctx.basepath);
```
tankakatan, line 35: As soon as we get rid of the `createBlockchain` function we'll be able to import the blockchain client directly from the module. The idea is to ultimately get rid of the `ctx` completely, since it is what makes all the modules coupled.

In other words, let's get rid of the `createBlockchain` function and import blockchain client directly everywhere we use it.

tankakatan, line 35: As soon as we get rid of the `createBlockchain` function we'll be able to import the blockchain client directly from the module. The idea is to ultimately get rid of the `ctx` completely, since it is what makes all the modules coupled.

In other words, let's get rid of the `createBlockchain` function and import blockchain client directly everywhere we use it.

tankakatan (changes requested):
Great work, thank you @germanviescas! Let's improve it a bit more and then merge it.

germanviescas:
Thanks @tankakatan! I rebased the branch, fixed conflicts, made the requested changes and re-tested.

tankakatan (approved):

PR #272, opened 2022-03-14
Make contracts administrable

elblockdude:


jensendarren:
@elblockdude looks good. When we move this to a separate npm library we should include tests for this.

PR #277, opened 2022-03-16
Move proxy server response html templates to separate files

brianunlam:


internal/explorer.z/public/identity.zhtml
```diff
@@ -32,7 +32,7 @@
 <table class="table table-bordered table-primary table-striped table-hover table-responsive">
```
brianunlam, line 32: this file should'n be included

tankakatan (approved):

PR #280, opened 2022-03-16
e2e using geth instead of ganache

diogosmendonca:
Makes e2e enviroment use geth instead of ganache. In this way, the e2e enviroment will be closer to production environment.


jensendarren (approved):

jensendarren:
@tankakatan are you ok if we merge this into `develop` ?

tankakatan:
> @tankakatan are you ok if we merge this into `develop` ?

Thanks @jensendarren and @diogosmendonca. Everything seems fine, just let's add new docker compose files instead of replacing existing ones right away, so that we would be able to compare and see any difference.

docker-compose.e2e.yaml
```diff
@@ -38,6 +38,7 @@ services:
       - shared_contracts_e2e:/build:rw
       - ./hardhat/contracts:/hardhat/contracts:ro
       - ./hardhat/scripts:/hardhat/scripts:ro
+      - ./hardhat/tasks:/hardhat/tasks:ro
```
tankakatan, line 41: I don't have such `file/directory`, does it belong to this PR?

tankakatan (approved):
All good, thanks. I've just added a few minor suggestions, please take a look.

docker-compose.e2e.yaml
diogosmendonca: I just get it from the develop branch, but I can remove it since will be add by a merge with develop.

docker-compose.e2e-geth.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - blockchain_data_e2e:/root/data:rw
+      - ./resources/geth_datadir/keystore:/root/data/keystore
+      - ./resources/geth-genesis.json:/root/geth-genesis.json
+      - ./resources/geth_passwords:/root/geth_passwords.txt
+      - ./resources/geth-init.sh:/root/geth-init.sh
+    environment:
+      - DOCKER=true
+    entrypoint: sh /root/geth-init.sh
+    command: |
+      --nodiscover --datadir /root/data --networkid 15 --http --http.api eth,net,web3,personal --http.addr 0.0.0.0 --http.port 7545 --http.corsdomain '*' --http.vhosts=blockchain_node --allow-insecure-unlock --unlock 8D94F8CAB38168AD421D2CF567c1da72c0ef8414 --password /root/geth_passwords.txt --mine
```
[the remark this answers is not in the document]
diogosmendonca: ok

docker-compose.e2e.yaml
jensendarren: Yea this is from another branch that is already merged into `develop`

PR #286, opened 2022-03-17
Establish docker test env and add storage tests

Koochr:


tests/docker/storage.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+describe('Storage upload/download', () => {
+    let id;
+    it('Should upload a file', async () => {
+        id = await uploadFile('Test string');
+        expect(id).toBeTruthy();
+    });
+
+    it('Should download an uploaded file', async () => {
+        await delay(120000); // Waiting for the file to be mined
```
tankakatan, line 17: Instead of just delay, let's poll `getFile` function once in 500ms or something until timeout. Then we can measure and see the actual time it takes to get mined.

tests/docker/storage.spec.js
```diff
@@ -0,0 +1,40 @@
...
+            {data: 'Test string 2'}
+        );
+        id = res.data.data;
+        expect(id).toBeTruthy();
+    });
+
+    it('Should download an uploaded file', async () => {
+        await delay(120000); // Waiting for the file to be mined
+        const res = await get(
+            `http://point_node:2468/v1/api/storage/getString/${id}`);
```
tankakatan, line 37: Nice, now we also need the same test for proxy

tankakatan (approved):
Thanks, great job, merging this. The below comments are for the next iteration.

PR #291, opened 2022-03-21
PN-94 Test blockchain integration.

germanviescas:


src/network/blockchain.js
```diff
@@ -109,7 +105,9 @@ blockchain.loadPointContract = async (
 };
 
 blockchain.loadIdentityContract = async () => {
-    const at = config.get('network.identity_contract_address');
+    const addressFromEnv = process.env.IDENTITY_CONTRACT_ADDRESS;
```
tankakatan, line 108: This functionality is already available in `config` module. Setting the `IDENTITY_CONTRACT_ADDRESS` variable already changes the config, no need to do this again. See `config/custom-environment-variables.yaml` file for more details.

tankakatan (commented):
The Identity contract address should be picked automatically. There's some problem with compose I think. Working on it already.

src/network/blockchain.js
germanviescas: I saw that, but it wasn't working as expected for the tests. And I think it makes sense, because the `IDENTITY_CONTRACT_ADDRESS` is set at runtime in _src/index.ts_, but that's not the entry point when we run _jest_; so, that env var is never set and the override in _config/custom-environment-variables.yaml_ has no effect.

PR #293, opened 2022-03-22
Feature/upgradable

diogosmendonca:
Allow upgradable contracts deployment. Use pointsocial for testing that is already adjusted to be a upgradable ZApp.


jensendarren (approved):
Looks good @diogosmendonca

PR #309, opened 2022-03-25
Not use stat function from fs-promises since it's broken for pkg library

brianunlam:
Related to this issue:

https://github.com/vercel/pkg/issues/1505


tankakatan (approved):

Koochr:
@brianunlam Please mind that `fs.promises.stat` is also used in `util/makeSurePathExists`

PR #317, opened 2022-03-30
PN-102 Initial implementation of RPC Provider endpoint.

germanviescas:
This PR implements support for all valid RPC methods, such as `eth_accounts`, `eth_getBalance`, etc.

The proxy intercepts requests to _https://web3.test/_ and maps them to _/v1/api/blockchain_. So, clients (such as Point SDK), can make a POST request to _https://web3.test/_ with the body: `{ method: string; params?: unknown[] }`.

To test it, I made requests from Postman, for example:
Screenshot from 2022-03-30 10-12-24

I also deployed a PointApp (blog.point) and interacted with it, to make sure nothing breaks.
Screenshot from 2022-03-30 11-14-19


tankakatan (approved):
Generally it is all fine.

The only thing that bothers me is that we may end up manually adding each Ethereum JSON RPC API method and then go on supporting all that.

What if we just use `web3.currentProvider.send` method instead, as it is mentioned [here](https://ethereum.stackexchange.com/a/3886) and [here](https://ethereum.stackexchange.com/a/68879).

I haven't checked it myself, but if it works, it will simplify a lot of things. What do you think @germanviescas?

germanviescas:
Absolutely @tankakatan, I did this temporarily because I thought we wanted to fully control the methods we expose for the initial test. I'll update the PR today.

germanviescas:
Updated.

tankakatan (approved):

PR #323, opened 2022-03-31
Fix point final and pointsocial feed

Koochr:


tankakatan (changes requested):
It looks like the current PR provides a sort of workaround to the broken functionality rather than a fix to it. The functions you removed from `renderer` still don't work, or are they?

tankakatan (approved):

PR #327, opened 2022-04-01
Validate chunks are uploaded to arweave

brianunlam:


src/db/models/chunk.js
```diff
@@ -13,6 +13,8 @@ export const CHUNK_UPLOAD_STATUS = {
     ENQUEUED: 'ENQUEUED',
     IN_PROGRESS: 'IN_PROGRESS',
     COMPLETED: 'COMPLETED',
+    VALIDATED: 'VALIDATED',
+    INVALID: 'INVALID',
```
Koochr, line 17: Also I don't see a reason to add one more status - we could have "validating" status with some timeout, reaching which should move chunk to "failed" state. In your logic it looks like after the first validation is failed (which is totally expected since mining can take some time), we consider chunk as "invalid" and don't check it anymore.

src/client/storage/uploader.ts
```diff
@@ -52,17 +60,20 @@ export const startChunkUpload = async (chunkId: string) => {
...
 
         // TODO: check status from bundler
-        if (response.data.status !== 'ok') {
+        if (response.data.status !== 'ok' || !response.data.txid) {
             throw new Error(`Chunk ${chunkId} uploading failed: arweave endpoint error: ${
                 JSON.stringify(response.data, null, 2)
             }`);
         }
 
+        const txid = response.data.txid;
```
Koochr, line 69: This seems to be a good idea to store txid, but how do we use it? Looks like we don't

PR #328, opened 2022-04-01
Validate chunks are uploaded to arweave

brianunlam:


src/client/storage/uploader.ts
```diff
@@ -90,3 +102,21 @@ export const uploadLoop = async () => {
...
+
+export const chunkValidatorLoop = async () => {
+
+    const allCompletedChunks: any[] = await Chunk.allBy('ul_status', CHUNK_UPLOAD_STATUS.VALIDATING, false);
+    allCompletedChunks.forEach(async (chunk) => {
+
+        try {
+            await arweave.transactions.getData(chunk.txid, {decode: true});
+            chunk.ul_status = CHUNK_UPLOAD_STATUS.COMPLETED;
+        } catch (error) {
```
Koochr, line 114: Why are you setting it to failed after the first failed attempt? We should implement some timeout or limit, after which it should be treated as failed. And, I think, this limit should be the DB field, same as we have for upload loop

src/client/storage/uploader.ts
```diff
@@ -52,17 +60,21 @@ export const startChunkUpload = async (chunkId: string) => {
         );
 
         // TODO: check status from bundler
```
Koochr, line 62: I think this TODO won't be actual after your fixes

Koochr (changes requested):

src/client/storage/uploader.ts
```diff
@@ -90,3 +100,21 @@ export const uploadLoop = async () => {
 
     setTimeout(uploadLoop, UPLOAD_LOOP_INTERVAL);
 };
+
+export const chunkValidatorLoop = async () => {
+
+    const allCompletedChunks: any[] = await Chunk.allBy('ul_status', CHUNK_UPLOAD_STATUS.VALIDATING, false);
+    allCompletedChunks.forEach(async (chunk) => {
```
tankakatan, line 107: I would propose to limit the array length here. If for some reason there is a hundred chunks to verify, we are going to request them all at the same time which may be not efficient.

src/client/storage/uploader.ts
```diff
@@ -6,6 +6,14 @@ import {promises as fs} from 'fs';
 import path from 'path';
 import config from 'config';
 import logger from '../../core/log';
+import Arweave from 'arweave';
+
+const arweave = Arweave.init({
```
tankakatan, line 11: We already have a set up arweave client in the `storage` module. Is it possible to reuse it by creating a sort of `getDataByTxId` in the `storage` and use it  within the `chunkValidatorLoop` instead?

tankakatan (commented):
Thanks, please see the comments below.

src/client/storage/uploader.ts
```diff
@@ -6,6 +6,11 @@ import {promises as fs} from 'fs';
 import path from 'path';
 import config from 'config';
 import logger from '../../core/log';
+import {storage} from './storage';
+import {eachLimit} from 'async';
+
+const MAX_UPLOAD_CONCURRENCY = 100;
+const MAX_VALIDATION_CONCURRENCY = 100;
```
tankakatan, line 13: This should not be hard-coded here. Please add to `config/default.yaml` and `config/custom_env_variables.yaml` instead. There is a property `storage.concurrent_upload_limit` already. I think the `CONCURRENT_UPLOAD_LIMIT` variable is also used for that.

tankakatan (approved):

PR #329, opened 2022-04-04
PN-99 Initial implementation of permissions for RPC methods.

germanviescas:
Following [EIP-2255](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2255.md), support for the following methods has been added:
- `wallet_getPermissions`
- `wallet_requestPermissions`

We now have a _preHandler_ that checks if the API request is for a restricted RPC method. Restricted methods are hard-coded in an array in src/api/middleware/rpc-method-permission.ts.

If the request is not for a restricted method, the _preHandler_ does nothing and the requests continues it's life-cycle.

If the request is for a restricted method:
1. it checks that the `request.headers.origin` is present, as this is needed to keep track of which dApp has access to which accounts.
    - If it is not present, it replies with a 400.
1. it fetches the `PermissionRecord` for that dApp + domain and checks if the RPC method being invoked is in the list of allowed methods or not.
    - If it is not, it replies with a 401.
    - If it is, the request continues and reaches the `BlockchainController`.

The way to revoke permissions is by sending a `wallet_requestPermissions` request with an empty array as `params`. This will revoke all permissions. At the moment, there's no support to remove permission for a single method.

Please find attached a Postman collection with a few requests that I used for testing purposes.
[RPC.postman_collection.json.zip](https://github.com/pointnetwork/pointnetwork/files/8411171/RPC.postman_collection.json.zip)


src/api/controllers/BlockchainController.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+    wallet_getPermissions: async (req: FastifyRequest) => {
+        const dappDomain = req.headers.origin;
+        if (!dappDomain) {
+            return {status: 400, result: {message: '`Origin` header is required.'}};
+        }
+
+        const address = blockchain.getOwner();
+        const permissions = await permissionStore.get(dappDomain, address);
+        return {status: 200, result: permissions || null};
```
germanviescas, line 44: the `result` is `permissions || null` because `permissions` could be `undefined`, and Fastify removes the key when the value is `undefined`. I thought it would cleaner for the client to always get a `permissions` key in the response, hence the `null`.

tankakatan (approved):

PR #359, opened 2022-04-13
Move utils

Koochr:
I've removed unused ones.
Also removed utils test, as functions tested there were not used anywhere else, so I removed them.
Non-used `threads/encrypt` and `threads/decrypt` files are using the old utils, but they are broken for a long time (they still import non-existing `defaultConfig.json`, so I left them as is.


Koochr:
Also , `hashFn` and `encode` functions were completely useless, as they were just calling functions from library (the first one also called `Buffer.from()` on a buffer), so I removed them and replaced usages with the built-in library functions

tankakatan:
> Also , `hashFn` and `encode` functions were completely useless, as they were just calling functions from library (the first one also called `Buffer.from()` on a buffer), so I removed them and replaced usages with the built-in library functions

Please undo this change. Both libraries should be encapsulated in utility functions because `hashFn` and `encode` implementation may change over time and in this case it needs to be changed everywhere.

tankakatan (approved):

PR #363, opened 2022-04-19
Fix/Proxy-Ws

tankakatan:


germanviescas:
How should we connect to the WS using the proxy?

For example, using Postman, without the proxy, I can establish a connection with the URL `localhost:2468`.

With the proxy, I tried using the URL `point` (which gets translated to _http://point/_), hoping the proxy would intercept this request and establish the WS connection, but it doesn't work.

Screenshot from 2022-04-19 14-35-25

tankakatan:
> 

Thanks @germanviescas, here's a snippet I've been using for testing.
```
(() => {
  const ws = new WebSocket('wss://point');
  ws.onerror = e => console.error('Custom ws error handler:', e, e.data);
  ws.onopen = () => {
    console.info('WS connection is open');
    ws.send(JSON.stringify({
      type: 'subscribeContractEvent',
      params: {
        contract: 'Hello',
        event: 'HelloWorld'
      }
    }));
  };
  ws.onmessage = (msg) => console.info('WS message:', msg);
  ws.onclose = (code) => console.info('Closed:', code);
})();
```

Also it the code we use to establish the ws connection in the sdk [here](https://github.com/pointnetwork/pointsdk/blob/main/src/pointsdk/index.ts#L219) should work.

So basically `host` doesn't matter at all, since all the requests get intercepted by the Proxy irregardless of requested url.

germanviescas:
Cool, thanks. I haven't figured out why it does not work in Postman, but at least I can see the connection is successfully established from Firefox when I use `const ws = new WebSocket('wss://point/')`.

I think [this line](https://github.com/pointnetwork/pointnetwork/blob/87d934cdb3f192cd73cc784c33eb318f1d3626ce/src/api/sockets/ZProxySocketController.js#L81) should be `this.wss.clients.forEach` instead of `this.wss.connections.forEach`, shouldn't it?

tankakatan:
> Cool, thanks. I haven't figured out why it does not work in Postman (probably an issue with the proxy config for ws protocol), but at least I can see the connection is successfully established from Firefox when I use `const ws = new WebSocket('wss://point/')`.
> 
> I think [this line](https://github.com/pointnetwork/pointnetwork/blob/87d934cdb3f192cd73cc784c33eb318f1d3626ce/src/api/sockets/ZProxySocketController.js#L81) should be `this.wss.clients.forEach` instead of `this.wss.connections.forEach`, shouldn't it?
> 
> Also in that block, instead of checking `client.state === 'open'`, we should check `client.readyState === WebSocket.OPEN` (or `=== 1` instead of the enum since it's a JS file).

Thanks @germanviescas, will take a look. I'll check the `client/connection` thing. You must be right, however I think it was previously working, or may be just not tested well. And I definitely will improve the state check, thanks.

PR #364, opened 2022-04-19
Allow URL rewriting and update explorer routes

diogosmendonca:
Allow to write routes that rewrite static urls making possible to refresh dynamic urls in SPAs.
For testing try to refresh the page on `https://point/identities/some_valid_identity`


jensendarren (approved):

jensendarren:
@diogosmendonca - looks good but one test is failing in `tests/docker/proxy.spec.js` - the failing test is `it('Should return https://point HTML')`. Can you take a look at that please?

PR #366, opened 2022-04-20
PN-120 Handle RPC requests through WS and REST

germanviescas:
URL to connect to WS: `wss://point/ws?token=POINTSDK_TOKEN`.

- The auth token is a hardcoded string for now, there's not signature checks, it's just a string comparison. The expected token is set in config `api.sdk_auth_key`.
- I introduced a dedicated route for WS connections: `/ws`. The reason is that I wanted to include a _preValidation_ hook that would only affect the WS connections and not the REST API requests. But if you think this could introduce breaking changes, I can leave it as it was and find another way of using the _preValidation_ hook.

RPC requests can be made by sending messages with the `type: 'rpc'`, for example:
```javascript
// Get account balance
{
    "type": "rpc",
    "method": "eth_getBalance",
    "params": ["0x916f8e7566dd63d7c444468cadea37e80f7f8048", "latest"]
}

// Initiate tx
{
    "type": "rpc",
    "method": "eth_sendTransaction",
    "params": [
        {
            "from": "0x916f8e7566dd63d7c444468cadea37e80f7f8048",
            "to": "0xd46e8dd67c5d32be8058bb8eb970870f07244567",
            "value": "0x1bc16d674ec80000",
            "gas": "0x76c0",
            "gasPrice": "0x9184e72a000",
            "data": "0xd46e8dd67c5d32be8d46e8dd67c5d32be8058bb8eb970870f072445675058bb8eb970870f072445675"
        }
    ]
}

// Confirm tx
{
    "type": "rpc",
    "method": "eth_confirmTransaction",
    "params": [
      {
        "reqId": "6bf07cdb83406acafec1ee9260292997915ad5fd5d5832ad3d1f5471fe8d5d93"
      }
    ]
}
````

Tx must be confirmed within 2 minutes, otherwise they will expire. This is set in config `rpc.send_transaction_timeout_secs`.

BTW: RPC requests can also be made to the REST API endpoint `/v1/api/blockchain`. In this case, use the header `Authorization: Bearer POINTSDK_TOKEN`.

### DIFF
--- a/src/api/sockets/ZProxySocketController.js
+++ b/src/api/sockets/ZProxySocketController.js
@@ -5,17 +5,20 @@ See client/proxy/index.js for usage details of ZProxy and setup of the WebSocket
 const logger = require('../../core/log');
 const log = logger.child({module: 'ZProxySocketController'});
 const blockchain = require('../../network/blockchain');
+const handleRPC = require('../../rpc/rpc-handlers').default;
 
 const SUBSCRIPTION_EVENT_TYPES = {
     CONFIRMATION: 'subscription_confirmation',
     CANCELLATION: 'subscription_cancellation',
     EVENT: 'subscription_event',
-    ERROR: 'subscription_error'
+    ERROR: 'subscription_error',
+    RPC: 'rpc_method_response'
 };
 
 const SUBSCRIPTION_REQUEST_TYPES = {
     SUBSCRIBE: 'subscribeContractEvent',
-    UNSUBSCRIBE: 'removeSubscriptionById'
+    UNSUBSCRIBE: 'removeSubscriptionById',
+    RPC: 'rpc'
 };
 
 class ZProxySocketController {
@@ -28,7 +31,7 @@ class ZProxySocketController {
     }
 
     init() {
-        this.ws.on('message', async (msg) => {
+        this.ws.on('message', async msg => {
             const {hostname} = this;
             const request = {
                 ...JSON.parse(msg),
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: src/api/sockets/ZProxySocketController.js [hunk 1]
```diff
@@ -78,14 +92,25 @@ class ZProxySocketController {
 
     pushToClients(msg) {
         if (this.wss) {
-            this.wss.connections.forEach(client => {
-                if (client.state === 'open') {
+            this.wss.clients.forEach(client => {
+                if (client.readyState === 1) {
```
tankakatan, line 96: Let's use named constant instead. It is more verbose.

src/client/proxy/handlers/common.ts
```diff
@@ -235,7 +235,25 @@ const attachCommonHandler = (server: FastifyInstance, ctx: any) => {
     const wsHandler = ({socket}: SocketStream, {hostname}: FastifyRequest) =>
         void new ZProxySocketController(ctx, socket, server.websocketServer, hostname);
 
-    server.route({method: 'GET', url: '*', handler, wsHandler});
+    // Handle websocket requests.
+    server.route({
+        method: 'GET',
+        url: '/ws',
```
tankakatan, line 241: I was under impression we decided to not use specific paths for ws connection?

tankakatan (approved):
Thanks @germanviescas, LGTM. I've added just a couple of minor comments we may address in the next PR.

src/client/proxy/handlers/common.ts
germanviescas: Yes, but as I noted in the description, when I started implementing it, I wanted to add the authorization as a _preValidation_ hook, and if we use the same handler for everythng, the _preValidation_ is going to affect everything and not just websocket.

But I can change it and find a different way to handle the authorization check, let me know.

src/api/sockets/ZProxySocketController.js
germanviescas: I assume you mean using the `WebSocket.OPEN` enum instead of the int `1`. But should we use an enum in a JS file?

PR #367, opened 2022-04-21
PN-122 Add github action to publish release on push to `master`.

germanviescas:
The idea is that with these changes, the flow to release Point Node would be:

- tag a commit on `develop` branch -> triggers github action to create a _DRAFT_ release
- test the _DRAFT_ release
- merge `develop` into `master`
- push `master` -> triggers github action to publish the _DRAFT_ release

### DIFF
--- a/.github/workflows/pkg.yaml
+++ b/.github/workflows/pkg.yaml
@@ -59,8 +59,8 @@ jobs:
       with:
         tag_name: ${{ steps.get_source_tag.outputs.SOURCE_TAG }}
         release_name: Release ${{ steps.get_source_tag.outputs.SOURCE_TAG }}
-        draft: false
-        prerelease: true
+        draft: true
+        prerelease: false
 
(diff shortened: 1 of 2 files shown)


.github/workflows/pkg.yaml
```diff
@@ -59,8 +59,8 @@ jobs:
       with:
         tag_name: ${{ steps.get_source_tag.outputs.SOURCE_TAG }}
         release_name: Release ${{ steps.get_source_tag.outputs.SOURCE_TAG }}
-        draft: false
-        prerelease: true
+        draft: true
+        prerelease: false
```
germanviescas, line 63: Instead of using a `prerelease`, I changed it to `draft`. The only reason is that I found an action to publish `draft` releases, but I couldn't find one to publish `prereleases`.

I think _drafts_ are ok for our use case, but please let me know if you have any concerns.

PR #368, opened 2022-04-22
Delegating deployment permission

diogosmendonca:
Allow to delegate permission to deploy to other addresses. CLI command in hardhat tasks. Point node validates if the user has permissions to deploy to an identity.

### DIFF
--- a/hardhat/hardhat.config.ts
+++ b/hardhat/hardhat.config.ts
@@ -11,10 +11,10 @@ import "./tasks/importer/identity";
 import "./tasks/importer/blog";
 import "./tasks/importer/pointSocial";
 import "./tasks/importer/sms";
-import "./tasks/importer/identity-update";
-import './tasks/importer/identity-add-deployer.ts';
-import './tasks/importer/identity-remove-deployer.ts';
-import './tasks/importer/identity-list-deployers.ts';
+import "./tasks/identity/identity-update-contract";
+import './tasks/identity/identity-add-deployer.ts';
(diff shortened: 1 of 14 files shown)


hardhat/hardhat.config.ts
```diff
@@ -67,12 +70,18 @@ const config: HardhatUserConfig = {
         development: {
             url: devaddress,
             accounts:
-            [privateKey],
+            [
+                privateKey,
+                '011967d88c6b79116bb879d4c2bc2c3caa23569edd85dfe0bc596846837bbc8e',
+                '22a316b515a68d4851087571bd5ff051f5ec3f13b28997fb80a8de055052514e'
+            ]
```
jensendarren, line 77: @diogosmendonca I dont think we need to add these two new private keys here because the first one is already set in the `privateKey` variable in line 39 and the second private key is the visitor node which we need to test by adding as a deployer so that should not be needed either. I've removed these keys and the process still works.

jensendarren (approved):

hardhat/hardhat.config.ts
jensendarren: @diogosmendonca I've removed the two private keys you added in my last commit. Do take a look and try this out when you are next around. Many thanks.

jensendarren (approved):
@diogosmendonca I've moved the identity tasks under the tasks/identity folder in Hardhat and made some other minor modifications. Please take a look. Thank you!

PR #373, opened 2022-04-26
PZ 146 Setup Dev environment and dclean script for Zapp dev env outside of Docker

diogosmendonca:
Scripts for automating and facilitate the usage of dev environment. Instruction for testing, run before all:
`source .bash_alias`
For creating a fresh new env (after clone for example, only needed in the first time). 
`point-dev-install`
After that it is possible to start all process for running point locally without docker, run:
`point-dev-start`
For stop all processes, run:
`point-dev-stop`
For clean up local data, run:
`point-dev-clean`

PS: Only tested in Linux.


jensendarren:
Feedback after testing this on Mac:

1. Can the `point-dev-install` also create and populate the `devlocal.yaml` and `visitlocal.yaml` files and have these ignored by git instead? This because the install script potentially makes local changes to these files anyway. Moreover in my case I need to hardcode my full path in the config to use `/Users/username` for some reason otherwise the key is not found (when using tilde ~). 
1. Can the `point-dev-start` script output the logs actually since if there is an error when starting the node its not returned to the user.
1. The `point-dev-start` script says `arlocal` starts when in fact it did not which caused the script to hang when starting the owner node later on.
1. The `point-dev-start` script cannot start using version `arlocal@1.1.21` and errors with `Unsupported engine error! Required: {"node":">=15.14.0"} Actual: {"npm":"8.5.5","node":"v14.17.5"}`. However when I install and run  arlocal@1.1.30 then it works!
1. The `point-dev-start` script does not continue after starting the owner node (even when the owner node is started correctly and running ok)
1. The `point-dev-start` script creates *.log files which I think should be ignored by git
1. killall node -v does not appear to kill all the node processes
1. In `devlocal.yaml` we should override the port numbers as shown [here](https://pointnetwork.github.io/docs/build-zapp-dev-environment-direct-install)

Thinking it might be better not to start the point node in the `point-dev-start` script and let the developer start the node as they like using the existing `point-dev` and `point-visit` alias commands. That way it solves the issue of the point node log output which I think should be easily visible in development environment.

diogosmendonca:
@jensendarren I changed the scripts to contemplate the requests made. 

- The logs are being showed in the console and not in log files.
- The config files are replaced every time that create or clean script runs, those files get back to be gitignored. 
- Start script only starts arlocal, ganache and deployer. The owner and visitor should be started by point-dev and point-visitor scripts. 
- Killall is sending SIGKILL instead of SIGTERM. 
- Arlocal version was updated to 1.1.30. 
- Fix ports in devlocal file.

jensendarren (approved):

PR #374, opened 2022-04-27
Implement multiple chains

Koochr:


config/default.yaml
```diff
@@ -66,7 +66,9 @@ wallet:
   passcode: BWVdmGdn8MeCWJxtcfM77eTMMxUELYL9 # TODO: remove
 network:
   contracts_path: contracts
-  web3: "http://ynet.point.space:44444"
+  web3:
+    ynet: "http://ynet.point.space:44444"
+    rinkeby: "https://rinkeby.infura.io/v3/400b1ad4183b442dbd9b0b5744b327aa"
```
Koochr, line 71: This is my account id, I've just created it. Probably, it could be replaced later

tankakatan (approved):

PR #386, opened 2022-05-03
Fallback to requesting chunks from bundler.

germanviescas:


config/default.yaml
```diff
@@ -34,7 +34,7 @@ storage:
   download_cache_path: client_storage_download_cache
   concurrent_upload_limit: 10
   concurrent_validation_limit: 10
-  upload_loop_interval: 500
+  upload_loop_interval: 2000
   request_timeout: 15000
```
brianunlam, line 38: I would increase this one, since today I've seen some errors because of this. 30000 would be good (request_timeout)

src/client/storage/index.js
```diff
@@ -68,32 +77,56 @@ const getChunk = async (chunkId, encoding = 'utf8', useCache = true) => {
...
+        log.warn(
+            {chunkId},
+            'Chunk not found in Arweave. Falling back to requesting it from bundler backup'
+        );
 
+        // TODO: refactor before mainnet!!!
+        // If we failed to retrieve the chunk from Arweave,
+        // retrieve it from the bundler's backup (S3).
+        try {
+            const url = `${config.get('storage.arweave_bundler_url')}/download/${chunkId}`;
```
Koochr, line 118: I'd move this up to constants same as other config values

src/client/storage/index.js
```diff
@@ -68,32 +77,56 @@ const getChunk = async (chunkId, encoding = 'utf8', useCache = true) => {
         log.debug({chunkId}, 'Graphql request success');
 
         for (const edge of queryResult.transactions.edges) {
```
Koochr, line 79: What I think, the flow is a bit redundant.
Here's how we download chunks from arweave: by chunk id we fetch graphql with tx ids array, than loop through this array trying to find data, which hash matches the id. In theory, there should not be any data with wrong id, but I think it can happen, maybe if data is broken. Also, it can happen for some reason that the array is empty.
So i think you should do the fallback s3 fetching not inside this loop - it actually doesn't make much sense, since s3 doesn't rely on tx id at all.
Instead, this fallback should be lower, in a catch block - so if arweave fails for any reason (graphql failed, or graphql res empty, or all of the nodes failed) - we fall back to s3

src/client/storage/index.js
```diff
@@ -68,32 +77,56 @@ const getChunk = async (chunkId, encoding = 'utf8', useCache = true) => {
...
             chunk.dl_status = CHUNK_DOWNLOAD_STATUS.COMPLETED;
             await chunk.save();
             return buf;
+        } catch (err) {
+            throw new Error('No matching hash found in Arweave nor bundler');
         }
-
-        throw new Error('No matching hash found in arweave');
     } catch (e) {
         log.error({chunkId, message: e.message, stack: e.stack}, 'Chunk download failed');
```
Koochr, line 131: Here's where the fallback should be, I think

germanviescas: I'll move it.

src/client/storage/index.js
germanviescas: I'll move it.

src/client/storage/index.js
germanviescas: I suggest we tackle this one separately. To have a hotfix out as soon as possible. There are many possibilities for improvement throughout the whole process.

config/default.yaml
germanviescas: done.

tankakatan (approved):

PR #394, opened 2022-05-09
Pz 108 enable zapp event subscription

emozilla:
This PR includes changes to allow applications to receive smart contract events via Websockets.

Some tweaked functionalities:

- WebSocket provider connection as default (http provider does not support subscriptions)
- Minor fix to hdwallet-provider (the .on method is not available)
- ZProxySocketController direct response to the opened socket (wss.clients or wss.connections were always undefined)


jensendarren:
@emozilla I am unable to deploy the Identity contract when testing this branch.  Getting `ProviderError: VM Exception while processing transaction: out of gas` can you check to see if you get the same. Many thanks.

jensendarren (approved):

PR #395, opened 2022-05-10
Implemented ZApp links as icons

emozilla:
This PR includes a small change to show deployed ZApps as icons

- It is necessary to build changes using npm run build inside of the folder /internal/explorer.point/
- The icons for each app comes from the favicon.ico of each ZApp, in the case that the favicon.ico is not available, a placeholder (currently the point logo) is loaded.

### DIFF
--- a/internal/explorer.point/src/custom.css
+++ b/internal/explorer.point/src/custom.css
@@ -15,3 +15,41 @@ a.nav-link.active {
     height: 1.6em; 
     margin-right: 0.8em;
 }
+
+.zapps {
+    display: grid;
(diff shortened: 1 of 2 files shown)


jensendarren:
Thank you @emozilla. Couple of observations:

- Currently we need to build and commit the public folder of the Point Explorer so can you do that and push an update here in this PR.
- Can you add an icon to the Email Zapp ( generic Email icon ), push a commit and PR to the Email Zapp repo so that I can see how that is picked up and displayed in the Point Explorer.

jensendarren (approved):

PR #405, opened 2022-05-13
Fix e2e test suite

germanviescas:
Two changes required to fix tests:
  - update tests to use `ynet_` prefix
  - properly handle the response from _Bundler_ when downloading a file, which is sent as a _stream_


germanviescas:
I will investigate why two tests failed here. They were all passing locally
tests

PR #406, opened 2022-05-13
Fix e2e test suite

germanviescas:
Two changes required to fix tests:
  - update tests to use ynet_ prefix
  - properly handle the response from Bundler when downloading a file,
  which is sent as a stream


src/client/storage/index.js
```diff
@@ -75,20 +75,14 @@ const getChunk = async (chunkId, encoding = 'utf8', useCache = true) => {
...
-            typeof data === 'object' ? Buffer.from(JSON.stringify(data)) : Buffer.from(data);
         await fs.writeFile(chunkPath, buf);
         chunk.size = buf.length;
-
-        const hash = hashFn(buf).toString('hex');
-        if (hash !== chunk.id) {
-            const errMsg = 'Chunk id and data do not match';
-            log.warn({chunkId: chunk.id, hash}, errMsg);
-            throw new Error(errMsg);
-        }
```
germanviescas, line 90: In order to make this handler a bit quicker and simpler, I think it's safe to skip this validation (lines 85:90) for content that comes from S3 since only _Bundler_ has write access to the bucket. We don't have the same risk as we do with Arweave, where anyone can upload random content and "fake" our tags.

brianunlam: I think that's ok having in mind that in my latest change to the bundler I'm validating the content on s3. On the other hand thanks to this part of the code we become aware about that something was wrong. I'm not 100%.

tests/docker/proxy-storage.spec.js
```diff
@@ -20,86 +21,95 @@ const httpsAgent = new HttpsAgent(httpAgentCfg);
...
+            expect(dirId).toBeTruthy();
+        },
+        TIMEOUTS.XL
+    );
+
+    it(
+        'Should download uploaded folder',
+        async () => {
+            expect.assertions(5);
+            await delay(5000);
```
brianunlam, line 103: it could be part of the TIMEOUTS constants

brianunlam (approved):

PR #412, opened 2022-05-17
add new private key env variable in hardhat config for ynet

jensendarren:


hardhat/hardhat.config.ts
```diff
@@ -38,6 +38,7 @@ task("accounts", "Prints the list of accounts", async (taskArgs, hre) => {
 });
 
 const privateKey = process.env.DEPLOYER_ACCOUNT || '0x011967d88c6b79116bb879d4c2bc2c3caa23569edd85dfe0bc596846837bbc8e';
+const ynetPrivateKey = process.env.DEPLOYER_ACCOUNT || 'YOU MUST SET THIS USING THE LOCAL ENVIRONMENT VARIABLE';
```
diogosmendonca, line 41: We can facilitate the management of this private key by converting and storing it together with the phrase in key.json. In the start script we could load it to the proper env variable.

PR #429, opened 2022-05-23
Add endpoint to resolve SNS and ENS domains

germanviescas:
`GET /v{n}/api/identity/:service/:domain` resolves a domain to it's owner address and also retrieves any content stored in the domain registry (if any).

Currently supported blockchains are:
  - Solana (SNS)
  - Ethereum (ENS)


config/default.yaml
```diff
@@ -73,9 +73,12 @@ network:
     rinkeby:
       type: "eth"
       address: "https://rinkeby.infura.io/v3/400b1ad4183b442dbd9b0b5744b327aa" # TODO: project ID
-#    solana_devnet:
-#      type: "solana"
-#      address: "https://api.devnet.solana.com"
+    # solana_devnet:
+    #   type: "solana"
+    #   address: "https://api.devnet.solana.com"
+    # solana:
```
germanviescas, line 79: I added Solana mainnnet details here (commented out) because it's much easier to test domain names in mainnet.

src/api/api_routes.js
```diff
@@ -15,6 +15,7 @@ module.exports = [
     ['POST', '/v1/api/identity/blockTimestamp', 'IdentityController@blockTimestamp'],
     ['POST', '/v1/api/identity/open', 'IdentityController@openLink'],
     ['POST', '/v1/api/identity/register', 'IdentityController@registerIdentity'],
+    ['GET', '/v1/api/identity/:service/:domain', 'IdentityController@resolveDomain'],
```
germanviescas, line 18: I think it makes sense to add this endpoint to the `identity` controller. But if you think there's a better place for it, please let me know.

src/api/controllers/IdentityController.js
```diff
@@ -284,6 +285,53 @@ class IdentityController extends PointSDKController {
 
         return this._response({eligibility, reason});
     }
+
+    async resolveDomain() {
+        const SUPPORTED_NAME_SERVICES = ['sns', 'ens'];
+        const SUPPORTED_TLD = ['.sol', '.eth'];
```
germanviescas, line 291: I was in doubt whether to hard-code these here or pass them in the _config_. I landed on this option thinking that these will rarely change and we probably won't need the flexibility of having different values in different environments. Also, _config_ is quite crowded as it is.

But if you think it's better to have them in _config_, please let me know.

src/api/controllers/IdentityController.js
```diff
@@ -284,6 +285,53 @@ class IdentityController extends PointSDKController {
...
+        if (!tld) {
+            const status = 400;
+            const errorMsg = `Unsupported TLD in "${domain}".`;
+            this.rep.status(status);
+            return this._status(status)._response({errorMsg});
+        }
+
+        try {
+            let registry;
+            switch (service) {
```
germanviescas, line 312: This `switch` is the kind of thing that we should probably avoid, having a thin abstraction over the providers. However, I'd like to make that change separately (I can create a subtask for that), in order to keep these changes as small and focused as possible.

src/network/providers/solana.ts
```diff
@@ -30,12 +34,40 @@ interface TransactionInstructionJSON {
...
+    ARWV = 'arwv'
+}
+
+type DomainContent = {
+    protocolType: Protocol | null;
+    decoded: string | null;
+    error?: Error;
+};
+
+type DomainRegistry = {
```
germanviescas, line 48: `web3.eth.ens` and `@solana/web3.js` return data in different formats. The goal here is to standardize the return value of the `resolveDomain` methods, so that, ultimately, the API always return the same response format, regardless of the blockchain.

This should also be clearer and easier to enforce when we add an abstraction for the providers.

src/network/providers/solana.ts
```diff
@@ -1,11 +1,15 @@
...
+import {getHashedName, getNameAccountKey, NameRegistryState} from '@solana/spl-name-service';
 import {mnemonicToSeedSync} from 'bip39';
 import {getSecretPhrase} from '../../wallet/keystore';
 import config from 'config';
 import axios from 'axios';
 const logger = require('../../core/log');
 const log = logger.child({module: 'SolanaProvider'});
 
+// Address of the `.sol` TLD
+const SOL_TLD_AUTHORITY = new web3.PublicKey('58PwtjSDuFHuUkYjH9BYnnQKHfwo9reZhC2zMJv9JPkx');
```
tankakatan, line 11: Minor: I would move the address to config. Even if the address is not going to change often, I think it would be easier to find it there. BTW is it the same address in both main net and test net?

src/api/controllers/IdentityController.js
tankakatan: It's fine to have it hard coded for now.

The thing I don't like is the service name that we expect to be passed. I think we should not rely on client code, and determine the service from the TLD itself instead, since there is no other way to use the services except requesting `sns` for `.sol` and `ens` for `.eth` domains.

tankakatan (approved):

src/api/controllers/IdentityController.js
germanviescas: Yeah, I wasn't sure about this. In theory, there could be other TLDs, and potentially some conflicts. For example, ethereum has `.test` in Ropsten. Solana could create a "domain program" with `.test` too.

But I guess the probability of collision is too small, and we are mostly interested in supporting `.sol` and `.eth`, so I'll make this change.

src/network/providers/solana.ts
germanviescas: Yes, it's the same address.

Ok, I'll move it to config and add it in the _default.yaml_ file.

PR #430, opened 2022-05-23
Refactor wallet module

Koochr:


src/network/providers/solana.ts
```diff
@@ -148,6 +142,66 @@ const solana = {
...
+    getBalance: async (network: string) => {
+        const provider = providers[network];
+        if (!provider) {
+            throw new Error(`Unknown network ${network}`);
+        }
+        const {connection, wallet} = provider;
+
+        return connection.getBalance(wallet.publicKey);
+    },
+    getSignaturesForAddress: async ({
```
tankakatan, line 183: The method names should be the same for all providers.

tankakatan (approved):

PR #438, opened 2022-05-30
mirror call on web2 redirection

nicobevilacqua:


src/client/proxy/handlers/mirror.ts
```diff
@@ -0,0 +1,45 @@
+import {parse} from 'query-string';
+import {FastifyRequest} from 'fastify';
+import blockchain from '../../../network/providers/ethereum';
+const {getJSON, getFileIdByPath} = require('../../storage');
+
+const CACHE_EXPIRATION = 1000 * 60 * 5; // FIVE MINUTES
```
tankakatan, line 6: Please move to `config/default.yaml`

tankakatan (approved):

src/client/proxy/handlers/mirror.ts
nicobevilacqua: done

jensendarren (approved):

PR #441, opened 2022-05-31
Add ability for proxy to resolve `.sol` domains

germanviescas:
This is one part of the "Support SNS and ENS" feature. Here, we're adding support for `.sol` domains.

Proxy can understand and resolve `.sol` domains in two different ways:

1. The `.sol` domain has all the required information in its registry, the identity, the routes ID and the root directory ID. This option will only be available for sites that do not have smart contracts.
2. The `.sol` domain is an alias to a `.point` domain. In this case, the Solana domain registry only stores the point alias, and Proxy uses it to retrieve all the required information, as with any other `.point` domain.

[Here's an example of the first scenario](https://drive.google.com/file/d/1-Eg0dxdJ7p2MSCr5lwdk_E5-VHY42Yze/view?usp=sharing).

[And here's an example of the second one](https://drive.google.com/file/d/1UW29EjnkIx_6WAbIUmQHVOmmJDUpWdOz/view?usp=sharing).

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,8 @@
       "#(.*)": "<rootDir>/core/$1"
     },
     "modulePathIgnorePatterns": [
-      "docker"
+      "docker",
+      "dist"
     ],
     "testEnvironment": "node",
     "testRunner": "jest-circus/runner",
--- a/src/client/proxy/handlers/common.ts
+++ b/src/client/proxy/handlers/common.ts
@@ -2,6 +2,7 @@
 import path from 'path';
 import {promises as fs} from 'fs';
 import {parse} from 'query-string';
+import axios from 'axios';
 import {makeSurePathExists, readFileByPath} from '../../../util';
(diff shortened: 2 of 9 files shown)


germanviescas:
The tests are failing with the usual timeout issue.
Screenshot from 2022-05-31 14-06-46


However, I also see this test failing, which doesn't seem related to timeouts, I'm marking the PR as **WIP** while I look into that
Screenshot from 2022-05-31 14-09-28
.

germanviescas:
Ok, the test has been broken not on this PR, but on the changes that added the "mirror" functionality (please see [here](https://github.com/pointnetwork/pointnetwork/blob/develop/src/client/proxy/handlers/common.ts#L247-L264)).

When trying to visit a domain other than `.point`, instead of a 404 as the test expects, we get a 200 with an HTML page that presents a link to web2.

germanviescas:
The test issues (200 vs 404) has been addressed in another PR. Proxy will now return 404 as the test expects. So no issues here.

PR #462, opened 2022-06-09
Making redirect using csrf

diogosmendonca:
Now for using web2 redirect the developer should use a link like this:
```
https://point/web2redirect?url=' + encodeURI('https://bounty.pointnetwork.io/')
                        + '&csrfToken=' + window.localStorage.getItem("csrf_token")
```
The point browser will open a confirmation page and when clicked the csrf will be checked against the host that called the web2redirect link (referrer). If csrf toke is ok the user will be redirected to web2.


totoyo1211 (approved):

jensendarren (approved):

jensendarren:
@diogosmendonca - this does not appear to be working as intended. If I change the `csrf` token to anything at all the link still opens the web2 site in my default browser. The common handler in the node should prevent this if the `csrf` token does not match.

jensendarren (approved):

PR #471, opened 2022-06-15
Extend IKV with @@copy_from_ikv= instruction

sergevar:
Allows to use `@@copy_from_ikv=` directive in certain scenarios.

This is only enabled if in `blockchain.getKeyValue()` the argument `followCopyFromIkv` is set to true.
By default it is false, and only set to true in two places: when asking for the root directory (`::rootDir`) and routes file (`zdns/routes`)

When this feature is activated, `blockchain.getKeyValue()` doesn't just return the key result, but checks if it is a special directive prefixed with `@@copy_from_ikv=`, for instance: `@@copy_from_ikv=anotheridentity:somekeythere`. Then it parses the directive and tries to obtain ikv value for `somekeythere` from identity `anotheridentity`. This is not recursive.

If this succeeds, it returns this value instead.

This is a useful helper until we have proper UI and backend for software packs. It would allow us to set root dir id and routes file id to certain values in, let's say, `blogsoftware.point`, and any user that has a handle registered, can just do these things to set up the blog functionality:

::rootDir: `@@copy_from_ikv=blogsoftware:v0.1.*/rootDir`
zdns/routes: `@@copy_from_ikv=blogsoftware:v0.1.*/routes`

It would also allow us to do versioning and update the functionality for all websites that point to blogsoftware, but only for minor releases of 0.1 version in this case.

### DIFF
--- a/src/network/providers/ethereum.js
+++ b/src/network/providers/ethereum.js
@@ -554,7 +554,7 @@ ethereum.getKeyValue = async (
         } else {
             // the format is:
             // @@copy_from_ikv=<identity>:<key_in_ikv>
-            const withoutPrefix = rootDirId.substring(copyFromIkv_prolog.length);
+            const withoutPrefix = value.substring(copyFromIkv_prolog.length);
 
             const copy_identity = withoutPrefix.split(':')[0];
             const copy_key = withoutPrefix.split(':').slice(1).join(':');
--- a/src/network/providers/ethereum.js
+++ b/src/network/providers/ethereum.js
@@ -15,6 +15,7 @@ const log = logger.child({module: 'EthereumProvider'});
 const {getNetworkPrivateKey, getNetworkAddress} = require('../../wallet/keystore');
 const {statAsync, resolveHome, compileAndSaveContract, escapeString} = require('../../util');
 const {createCache} = require('../../util/cache');
+const {copy} = require("fs-extra");
 
 function isRetryableError({message}) {
     for (const code in retryableErrors) {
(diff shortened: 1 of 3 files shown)


src/network/providers/ethereum.js
```diff
@@ -540,8 +540,32 @@ ethereum.getKeyValue = async (
...
     key,
     version = 'latest',
-    versionSearchStrategy = 'exact'
+    versionSearchStrategy = 'exact',
+    followCopyFromIkv = false
 ) => {
+    // Process @@copy_from_ikv instruction if followCopyFromIkv is set to true
+    if (followCopyFromIkv) {
+        // self invoke to get the value first but without redirection
+        const value = ethereum.getKeyValue(identity, key, version, versionSearchStrategy, false);
```
germanviescas, line 549: You'll probably want to `await` this one.

sergevar: yep, thanks!

src/network/providers/ethereum.js
```diff
@@ -540,8 +540,32 @@ ethereum.getKeyValue = async (
...
+        // self invoke to get the value first but without redirection
+        const value = await ethereum.getKeyValue(identity, key, version, versionSearchStrategy, false);
+
+        const copyFromIkv_prolog = '@@copy_from_ikv=';
+        if (! value.startsWith(copyFromIkv_prolog)) {
+            return value; // no redirection
+        } else {
+            // the format is:
+            // @@copy_from_ikv=<identity>:<key_in_ikv>
+            const withoutPrefix = rootDirId.substring(copyFromIkv_prolog.length);
```
germanviescas, line 557: Where is this `rootDirId` coming from? I cannot find it defined. I think this should be the `value` we get above in line 549.

tankakatan (approved):

PR #473, opened 2022-06-15
Feature/wallet tokens

Koochr:


src/api/api_routes.js
```diff
@@ -32,6 +32,9 @@ module.exports = [
     ['GET', '/v1/api/wallet/hash', 'WalletController@hash'],
     ['GET', '/v1/api/wallet/address', 'WalletController@address'],
     ['GET', '/v1/api/wallet/getWalletInfo', 'WalletController@getWalletInfo'],
+    ['GET', '/v1/api/wallet/getTokenBalances', 'WalletController@getTokenBalances'],
```
tankakatan, line 35: We don't need this endpoints in the API server, they should be handled by the Proxy api.

internal/explorer.point/src/pages/Wallet.jsx
```diff
@@ -124,38 +165,112 @@ export default function Wallet() {
         );
     };
 
+    const renderToken = (token, network) => {
```
tankakatan, line 168: I don't understand why we keep avoiding react app best practices so hard. What is the reason for this function to be declared in the body of another component instead of being a separate functional one? What are the advantages such an approach provides?

src/api/controllers/WalletController.js
```diff
@@ -1,14 +1,17 @@
 const PointSDKController = require('./PointSDKController');
 const ethereumjs = require('ethereumjs-util');
-const blockchain = require('../../network/providers/ethereum');
-const {getNetworkPublicKey} = require('../../wallet/keystore');
+const ethereum = require('../../network/providers/ethereum');
```
tankakatan, line 3: I haven't found a place where we check the WalletController methods are called by a privileged host `point`.

tankakatan (changes requested):

tankakatan (approved):

PR #477, opened 2022-06-21
PN-175 Support CORS in proxy server

germanviescas:
Supporting CORS in proxy is needed to get assets from the `mirror`. Please refer to [the ticket](https://point-labs.atlassian.net/browse/PN-175) for further details.

cors

CORS is only enabled for requests to mirror.point.

These changes only handle _simple_ requests. We may need to also support _preflight_ requests, but since I haven't came across them just yet, we could add support for them later on (however, if you want me to do it now, let me know).

### DIFF
--- a/src/client/proxy/middleware/cors.ts
+++ b/src/client/proxy/middleware/cors.ts
@@ -1,0 +1,18 @@
+import {preHandlerHookHandler} from 'fastify';
+
+/**
+ * Allows Cross-Origin Requests to `https://mirror.point`.
+ */
+export const cors: preHandlerHookHandler = (req, res, done) => {
+    const protocol = req.protocol;
+    const {host} = req.urlData();
+
+    if (`${protocol}://${host}` === 'https://mirror.point') {
+        // Requests that are a result of a redirection during a CORS request
+        // have the `Origin` header set to `null`.
+        // Hence, we allow all origins (*) for `https://mirror.point`.
+        res.header('Access-Control-Allow-Origin', '*');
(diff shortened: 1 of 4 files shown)


tankakatan (changes requested):
Cross origin requests should not be allowed by default

src/client/proxy/middleware/cors.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Allows Cross-Origin Requests to `https://mirror.point`.
+ */
+export const cors: preHandlerHookHandler = (req, res, done) => {
+    const protocol = req.protocol;
+    const {host} = req.urlData();
+
+    if (`${protocol}://${host}` === 'https://mirror.point') {
+        // Requests that are a result of a redirection during a CORS request
+        // have the `Origin` header set to `null`.
+        // Hence, we allow all origins (*) for `https://mirror.point`.
```
germanviescas, line 13: This is intentional behaviour:

> If a cross-origin resource redirects to another resource at a new origin, the browser will set the value of the Origin header to null after redirecting. This prevents additional confused deputy attacks [...]

If you don't want the wildcard, please let me know and I'll see what we can do about it. The only thing that comes to mind is not redirecting, but perhaps there are other solutions I'm not aware of.

tankakatan (dismissed):
Please update the PR

germanviescas:
Conflicts with `develop` solved.

tankakatan (approved):

PR #486, opened 2022-06-28
Fix explorer to work with new point engine and sdk

diogosmendonca:
Changes performed: 
1. Fix owner identification because of address case in the version of point engine.
2. Change contract.send to contract.call to work with new version of point SDK.
3. Improve error message presentation.


jensendarren (approved):

jensendarren:
Thank you @diogosmendonca. This can be merged in when the latest Point SDK is released.

PR #487, opened 2022-06-29
Feature/remove ctx

Koochr:
- removed ctx object
- moved `csrf_tokens` from ctx to separate file
- remove `deployerProgress`
- refactored some redundant classes, which only contained initialization functions, to plain functions
- splitted api server to 2 files for the same purpose as it was done with proxy - for adding unit tests
- rewritten storage and some other files from js to ts
- fix zappdev env


docker-compose.zappdev.yaml
```diff
@@ -36,9 +36,10 @@ services:
       - shared_contracts_zappdev:/build:rw
 
   point_node:
-    extends:
-      file: docker-compose.base.yaml
-      service: point_node
+    build:
```
tankakatan, line 39: This will slow down the `zappdev` env, what's the advantage of this change?

src/client/storage/index.ts
```diff
@@ -337,7 +162,7 @@ const uploadDir = async dirPath => {
     const files = await fs.readdir(dirPath);
     const dirInfo = {
         type: 'dir',
-        files: []
+        files: [] as any[] // TODO: any
```
tankakatan, line 165: Let's define the file type right away, it seems to be already clear form the below structure (line 176).

src/client/storage/uploadChunk/uploadChunk.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    log.debug({chunkId}, 'Enqueuing chunk for upload');
+    chunk.ul_status = CHUNK_UPLOAD_STATUS.ENQUEUED;
+    await chunk.save();
+
+    let uploaded = false;
+
+    const check = async () => {
+        // TODO: any
+        const updatedChunk: any = await Chunk.find(chunkId);
```
tankakatan, line 26: I think the chunk structure is more or less stable, let's also create a type for it.

src/api/server.ts
```diff
@@ -0,0 +1,74 @@
+import fastify, {HTTPMethods} from 'fastify';
+import logger from '../core/log';
+import fastifyWs from 'fastify-websocket';
+import {transformErrorResp} from '../errors';
+import identityMdw from './middleware/identity';
+const ws_routes = require('./ws_routes');
+const api_routes = require('./api_routes');
```
tankakatan, line 7: Ok, since we already have started refactoring API server, let's also get rid of the `api_routes` and `ws_routes` files. First of all the handlers themselves should be functions, not classes. It makes no sense to instantiate ones just to simply handle a requests. Given that we don't need to have the mapping between urls and class methods, we can simply attach the handlers here the same way we do in proxy.

tankakatan (changes requested):
Generally it looks much better now. Please review the below comments and we'll be ready to merge.

docker-compose.zappdev.yaml
Koochr: I just copied it from the `test` compose file, as with the original config it was starting outdated point node image. However, I'll roll it back so guys from Darren's team can decide themselves

src/api/server.ts
Koochr: This should be quite a large change, so, as I proposed earlier, let's leave it for the next PR

docker-compose.zappdev.yaml
tankakatan: > I just copied it from the `test` compose file, as with the original config it was starting outdated point node image. However, I'll roll it back so guys from Darren's team can decide themselves

Yeah, it makes sense to build the image in the test compose because we always want to test it against the latest changes. But it case of `zappdev` compose it is a bit different, we need to focus of the `zapps` development, so re-building the image will probably be very annoying.

src/client/storage/uploadChunk/uploadChunk.ts
```diff
@@ -0,0 +1,54 @@
+import {delay, hashFn} from '../../../util';
```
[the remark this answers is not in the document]
Koochr: One change here. I'll make prefix `zappdev` instead of just `dev`, otherwise it will seem confusing - we're not using these files in a dev env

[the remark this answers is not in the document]
tankakatan: Ok, makes sense.

src/client/storage/uploadChunk/uploadChunk.ts
Koochr: Let's also leave it for the next PR. `Chunk` is a model, and writing type for it requires rewriting all our `db` folder in typescript.

tankakatan (approved):

PR #494, opened 2022-07-05
Create an abstraction to unify blockchain providers

germanviescas:
This is the initial step towards an abstraction to consolidate Blockchain Providers functionality, providing a unique `interface` that all Providers must adhere to, and a single _exportable_ Blockchain Provider that internally chooses which specific Provider (Ethereum or Solana for now) to call.

Anyone wishing to interact with blockchain just needs to import the _main_ provider from `src/network/providers/provider.ts` and call any of the methods in the `Provider` interface. By passing a `chainId`, this provider will choose the correct chain to use.

Hopefully, the changes to `src/rpc/rpc-handlers.ts` help illustrate the usage and the benefits of this approach.

Some things I'd like to point out:

### CONFIG/ENUMS

I added a `chain_id` field to each chain. We could use the keys we already have there to uniquely identified each chain, but I think it would be better for us to stick to generally known chain IDs, such as **4** instead of **rinkeby**, for example. The same is reflected in a `ChainId` enum that I introduced.

### TESTS

Some of the tests I added are not proper unit tests in that they connect to blockchain providers. Only read operations, in order not to run into issues with lack of funds.

On one hand, I don't think it's a great idea for tests to depend on external services. On the other hand, they provide some useful insight, event if they fail for external reasons, that's valuable information. So I'm on the fence about leaving them as they are, or mocking the providers.

Two test cases are commented out, both regarding Name Services:
* The ENS one is commented out because it needs to connect to Rinkeby and we're regularly hitting "too many requests" nowadays. We can uncomment it once we have a premium Infura subscription.
* The SNS one is commented out because it need to connect to mainnet, and we may not want to add a mainnet provider yet.

### TYPE DECLARATIONS

Two packages have no type declaration files available:
* web3-provider-engine (it's not actively maintained)
* @ensdomains/eth-ens-namehash

So I had to ts-ignore those imports.

There's still a lot to cover, but I think this introduces the main design to make the rest of the migration.

One thing that this design doesn't include yet, is how to handle WS connections, because I figured it would be better to decide how to handle them when we tackle that specific functionality.


src/network/providers/ethereum-provider.ts
```diff
@@ -0,0 +1,214 @@
+import {ethers} from 'ethers';
```
germanviescas, line 1: This file will eventually replace `./ethereum.js`.
Keeping both, while a bit untidy, allow us to make a gradual migration.

src/network/providers/ethereum-provider.ts
```diff
@@ -0,0 +1,214 @@
...
+        return resp;
+    }
+
+    public async requestAccounts(chainId: ChainId, id: number): Promise<JsonRpcResponse> {
+        const accts = await this.send(chainId, {id, method: 'eth_accounts'});
+        return accts;
+    }
+
+    public async getBlockNumber(chainId: ChainId): Promise<number> {
+        const provider = this.getChainProvider(chainId, 'ethers');
```
germanviescas, line 208: We could take advantage of the migration process to use `ethers` instead of `web3.js`. This used to be done with `web3.js` and is now done with `ethers`.

src/network/providers/ethereum-provider.ts
```diff
@@ -0,0 +1,214 @@
...
+        const n = await provider.getBlockNumber();
```
germanviescas, line 209: We could take advantage of the migration process to use `ethers` instead of `web3.js`. This used to be done with `web3.js` and is now done with `ethers`.

src/network/providers/ethereum.js
```diff
@@ -113,8 +113,9 @@ const getEthers = chain => {
     }
 
     if (!ethersProviders[chain]) {
-        const url = networks[chain].address;
-        ethersProviders[chain] = new ethers.providers.JsonRpcProvider(url);
+        const host = networks[chain].address;
+        const protocol = networks[chain].tls ? 'https' : 'http';
+        ethersProviders[chain] = new ethers.providers.JsonRpcProvider(`${protocol}://${host}`);
```
germanviescas, line 118: This fixes an actual bug that was introduced when removing protocols from config. I'm including the fix here and not as a hotfix because it only affects ENS, which is pretty much an unused functionality.

src/network/providers/provider.ts
```diff
@@ -0,0 +1,141 @@
...
+    resolveDomain(chainId: ChainId, domain: string): Promise<DomainRegistry>;
+    setDomainContent(
+        chainId: ChainId,
+        domain: string,
+        data: string
+    ): Promise<Record<string, string|number|boolean>>;
+    send(chainId: ChainId, data: RPCRequest): Promise<JsonRpcResponse>;
+    sendTransaction(chainId: ChainId, reqId: number, txParams: unknown[]): Promise<JsonRpcResponse>;
+    requestAccounts(chainId: ChainId, reqId: number): Promise<JsonRpcResponse>;
+    getBlockNumber?(chainId: ChainId): Promise<number>;
```
germanviescas, line 37: Methods that not all Providers implement are marked as optional. Then we check if it's implemented (comment below).

src/network/providers/provider.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    public async requestAccounts(chainId: ChainId, reqId: number): Promise<JsonRpcResponse> {
+        const accts = await this.getProvider(chainId).requestAccounts(chainId, reqId);
+        return accts;
+    }
+
+    public async getBlockNumber(chainId: ChainId): Promise<number> {
+        const provider = this.getProvider(chainId);
+        if (!provider.getBlockNumber) {
+            throw this.notImplementedErr(chainId, 'getBlockNumber');
```
germanviescas, line 133: Since `getBlockNumber` is optional, meaning not all Providers will implement it. We check here and throw a clear error.

src/network/providers/provider.ts
```diff
@@ -0,0 +1,141 @@
...
+    public async sendTransaction(
+        chainId: ChainId,
+        reqId: number,
+        txParams: unknown[]
+    ): Promise<JsonRpcResponse> {
+        const resp = await this.getProvider(chainId).sendTransaction(chainId, reqId, txParams);
+        return resp;
+    }
+
+    public async requestAccounts(chainId: ChainId, reqId: number): Promise<JsonRpcResponse> {
```
germanviescas, line 125: All these methods could be surrounded in try/catch, but I think it's better for the caller to handle errors. We are already have some duplication when it comes to handling and logging errors.

src/network/providers/provider.ts
```diff
@@ -0,0 +1,141 @@
...
+    return Object.values(ChainId).includes(chainId);
+}
+
+/**
+ * BlockchainProvider is the abstraction that consolidates and exposes
+ * blockchain functionality and provides a unified solution for working
+ * with different chains such as Ethereum and Solana.
+ */
+class BlockchainProvider implements Provider {
+    private providers: {[key in Blockchain]?: Provider} = {};
```
germanviescas, line 65: `providers` looks something like:

```
providers: {
    ethereum: EthereumProvider,
    solana: SolanaProvider,
}
```

It doesn't handle chains within the same blockchain protocol, that's left to the specific providers. For example, Mainnet, Rinkeby, Ropsten, are handled within the `EthereumProvider`.

src/network/providers/solana-provider.ts
```diff
@@ -0,0 +1,283 @@
+import {
```
germanviescas, line 1: This file will eventually replace `./solana.ts`.

src/network/providers/solana-provider.ts
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * Singleton class to make calls and send transactions to Solana blockchain.
+ * With potential support for multiple chains (mainnet, devnet, etc.).
+ */
+class SolanaProvider implements Provider {
+    private static instance: SolanaProvider;
+
+    // Lazily instantiated web3 providers to interact with different Solana chains.
+    private chains: {[key in ChainId]?: SolanaConnection} = {};
```
germanviescas, line 72: Unlike `EthereumProvider`, we use a single library here so `chains` is simpler:
```
chains: {
    devnet: SolanaWeb3Provider,
    'mainnet-beta': SolanaWeb3Provider,
}
```

src/network/providers/provider.ts
```diff
@@ -0,0 +1,141 @@
...
+};
+
+export const tldToChain: Record<string, ChainId> = {
+    '.sol': ChainId.SolanaMainnet,
+    '.eth': ChainId.EthereumRinkeby
+};
+
+// Type guard to determine if a given string|number is a valid ChainId
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function isChainId(chainId: any): chainId is ChainId {
```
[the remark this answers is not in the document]
tankakatan: I think we can narrow it a bit if we declare it like `string | number`. Or are there other cases I don't know of?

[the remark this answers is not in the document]
germanviescas: Sounds good. I think I'll have to add `boolean` too, as the config has some. But I'll double check and update the PR.

germanviescas:
Marked as a **WIP** because some tests in the files below rely on blockchain calls and we should avoid that in unit tests.
- src/rpc/rpc-handlers.test.ts
- src/network/providers/provider.test.ts

PR #515, opened 2022-07-20
Add eslint curly rule

Koochr:


brianunlam:
Nice, I didn't like that one-liner if either.

tankakatan (approved):

PR #520, opened 2022-07-21
Feature/api tests rebased

tankakatan:


germanviescas (commented):
I wonder why we have changes in `internal/explorer.point/public/index.js`, considering we don't have any changes in **src** (**public** is the compiled JS).

Maybe we should undo these changes?

PR #531, opened 2022-07-26
PN-195 Support SNS and ENS for identity resolution in Point Network

germanviescas:
These changes make it possible for addresses that own either an SNS or ENS domain to use such domain as their identity, without needing to register a Point Identity.

I made some changes in _explorer.point_ to accomodate for SNS and ENS domains. I think everything is working but would appreciate a fresh set of eyes if anybody is available to test.

As we discussed, there's likely some breaking changes in some Point Apps. One of the main issues potentially being encryption using other users' public keys. But other things may break, for example, the _wallet_ methods in PointSDK only support Ethereum, so, depending on which Apps use them, and how they use them, there might be issues.


src/api/controllers/IdentityController.js
```diff
@@ -118,24 +177,63 @@ class IdentityController extends PointSDKController {
...
+
+        const identity = await getIdentity({[key]: owner, targetOnly});
         return this._response({identity});
     }
 
     async publicKeyByIdentity() {
         const identity = this.req.params.identity;
+        if (identity.endsWith('.sol') || identity.endsWith('.eth')) {
+            this.rep.status(422);
+            return this._status(422)._response('This endpoint only supports Point identities.');
```
germanviescas, line 234: For Ethereum, I don't think it's possible to get a Public Key from an Address. Addresses are derived from Public Keys but have "less data", so going the other way is not possible. I assume that's the reason we keep the Public Keys in the Identity Contract.

For Solana, I understand the Public Key and the Address are the same thing, so we shouldn't have this problem. I would need to confirm it though.

When the identity is an ENS domain, this is going to be a problem for dApps that need the Public Key. For example _email.point_ and most likely _drive.point_.

If they only need the public key of the "logged in" user, we can get it because we have the secret phrase. But, if they need one from another ENS identity, to encrypt an email (or a shared file) for example, we won't be able to get it this way.

For SNS identities, we need to check if encrypting with Solana Public Keys works as expected.

src/api/controllers/IdentityController.js
```diff
@@ -110,6 +110,65 @@ function getHashedMessage(identity, owner, type) {
...
+ *
+ * Identity can be:
+ *   - a Solana (SNS) domain (`.sol`)
+ *   - an Ethereum (ENS) domain (`.eth`)
+ *   - a Point identity (registered in Point Identity Contract)
+ *
+ * The priority to choose an identity (should multiple be available) is:
+ *   - SNS
+ *   - ENS
+ *   - Point
```
germanviescas, line 125: With this approach (looking first in SNS, then ENS and lastly Point), if an existing Point Identity happens to own an SNS or ENS domain with the same address, the next time the user access Point Network, they'll see their identity changed to the SNS or ENS domain.

How do we plan to handle these cases?

For example, this is the wallet page of an existing Point Identity `ynet_dell.point`, that also owns `pointnetwork.eth`:
wallet_existing_identity

src/network/providers/ethereum.js
```diff
@@ -912,6 +934,20 @@ ethereum.resolveDomain = async (domainName, network = 'rinkeby') => {
     return {owner, content};
 };
 
+/**
+ * Retrieve ethereum domain (ENS) owned by `address`.
+ *
+ * An ENS name requries additional configuration to setup a reverse record,
+ * they are not automatically set up.
```
germanviescas, line 941: ENS reverse record must be manually set up by the domain owner for this to work.
ens_reverse-record

germanviescas:
We should update the identity page if the identity is an SNS or ENS domain, shouldn't we?

IKV and Deployers wouldn't make sense in such cases, would they?

eth_identity_page

germanviescas:
For example in the navbar of the _explorer_, we must be calling `ownerToIdentity` to show the identity link (last one).

In this context, `owner` is an Ethereum address. This is a problem, because if the identity is a `.sol` one, the "owner" is a Solana account, and we will miss it. It's simple enough to fix when we are looking for the identity of the "logged in" user, the `isIdentityRegisterd` endpoint now returns the `identity`, we could use that, or introduce a new endpoint with a clearer name.

The problem could be when `ownerToIdentity` is used for "external" users. In those cases, we don't have a way to go from an Ethereum address to a Solana one, to check if this "external user" is using a Solana identity. Hopefully, there is no such usage of `ownerToIdentity`, but need to check carefully.

src/api/controllers/IdentityController.js
tankakatan: Thanks @germanviescas, in order to send funds we use address and chain id, so no problem here.

In the UI for each network we can show the appropriate handle or the address relevant to the chain if the handle is not registered.

So let's say you have an address like `0x123...` and a corresponding Solana address `4567...`, which is the owner of `ynet_foo.point` identity and `foo.sol` Solana devnet domain respectively. In the wallet we should see:

| Currency   |      Address      |  Balance |
|----------|:-------------:|------:|
| Point |  ynet_foo.point | ... |
| yPoint |    ynet_foo.point   |   ... |
| rinkeby | 0x123... |    ... |
| devSol | foo.sol |    ... |
| Sol | 4567... |    ... |

In other words, show handles only for those chains where they exist, and the addresses for the rest ones. Does this answer your question?

germanviescas:
Probably not a big issue, but in _My Identities_, we won't be able to list SNS and ENS identities.

src/api/controllers/IdentityController.js
tankakatan: Yes, I think this method should only be used for Point identities. Let's throw an error in case `.sol` or `.eth` identity is passed for now.

src/network/providers/ethereum.js
```diff
@@ -578,17 +591,26 @@ ethereum.getKeyValue = async (
...
         } else {
             // the format is:
             // @@copy_from_ikv=<identity>:<key_in_ikv>
             const withoutPrefix = value.substring(copyFromIkv_prolog.length);
 
             const copy_identity = withoutPrefix.split(':')[0];
-            const copy_key = withoutPrefix.split(':').slice(1).join(':');
-
-            const value = await ethereum.getKeyValue(copy_identity, copy_key, 'latest', 'exact', true);
+            const copy_key = withoutPrefix
```
tankakatan, line 602: Is there a way to avoid changes like that? I assume it was made automatically, right? I think in this case the changed string was less than 100 chars long. If it's true, then the change makes no sense and only complicates reviews.

tankakatan (dismissed):

src/network/providers/ethereum.js
germanviescas: Yes, it was made automatically by running prettier as per the _.prettierrc.js_ config in the repo.

In my opinion, the best way to avoid is for everybody to run prettier on their changes, it's quite simple to set up "format on save" in any editor.

A second option would be to remove _.prettierrc.js_ and not use prettier at all.

A third option would be to leave things like they are and for me to turn off prettier locally. I just don't like to having a _.prettierrc.js_ and not using it. But I can do it if that's the preference.

src/api/controllers/IdentityController.js
```diff
@@ -110,6 +110,97 @@ function getHashedMessage(identity, owner, type) {
...
+ * If you are interested in one source of identity only, set a `targetOnly`:
+ *   - `targetOnly === 'solana'` -> identity is only searched for in SNS.
+ *   - `targetOnly === 'ethereum'` -> identity is only searched for in ENS and Point Identity Contract.
+ */
+async function getIdentity(
+    {solAddress, ethAddress, targetOnly, solNetwork, ethNetwork} = {
+        solAddress: '',
+        ethAddress: '',
+        targetOnly: '',
+        solNetwork: 'solana',
```
germanviescas, line 136: What should we do when it comes to connecting to Solana?
 
For now, the default here is to use solana mainnet, but for that work we would need to uncomment the solana provider information in config/default.yaml. The problem is that we would be using a mainnet in our ynet, and I don't know if we want to do that.

I tried using Bonfida (for SNS stuff) with solana_devnet and solana_testnet but it doesn't work. I asked them some time ago and they said they only supported mainnet.

internal/explorer.point/src/context/AppContext.js
```diff
@@ -18,22 +18,36 @@ const AppContext = createContext(defaultContext);
...
-            const {
-                data: { address },
-            } = await window.point.wallet.address();
-            const {
-                data: { identity },
-            } = await window.point.identity.ownerToIdentity({ owner: address });
-
-            setWalletIdentity(identity);
+            let identityData = {};
+            if (window.point.identity.me) {
```
germanviescas, line 32: This is a new method I added in a PR in **PointSDK**. It falls back to a _fetch_ request directly to the API so as to not couple both set of changes. This will work even if we don't deploy the changes in PointSDK.

germanviescas:
Here are some screenshots of _explorer_ with the three different kind of identities.

**POINT**
identity_point

**SOL**
identity_sol

**ETH**
identity_eth

germanviescas:
@tankakatan, now that we look for identities in SNS and ENS, identity resolution has gotten slower. You will notice that simply loading https://point takes 2-3 seconds.

Shall we implement a cache? If so, should we do it in this PR so that we don't deploy this "slower" version? Or is it ok to do it afterwards?

tankakatan (dismissed):

germanviescas:
With the latest change, if a domain exists in SNS or ENS, we show that instead of the address in the wallet page:

Screenshot from 2022-08-04 16-47-58

Screenshot from 2022-08-04 16-49-45

In the second picture, the user does not have a Point Identity, only a `.sol` domain. As a result, the **Point** and **xPoint** addresses are shown as **N/A**. Should we leave it like that or show the ethereum address instead?

tankakatan (approved):

PR #533, opened 2022-07-29
v0.2.4 release merge to main

jensendarren:


germanviescas (approved):
Merging to `main` will indeed trigger the GitHub action to publish the latest draft release. In this case, since we don't have any draft releases, this should be safe. The workflow does not fail if no draft releases are found, it does nothing.

PR #535, opened 2022-08-01
Replace udp stream when it fails

brianunlam:


src/util/createUdpStream.ts
```diff
@@ -0,0 +1,26 @@
...
+    });
+}
+
+export function createUdpStream(options: { address: string; port: number }) {
+    let writable = createWritable(options);
+    const erroHandler = () => {
+        writable = createWritable(options);
+        writable.on('error', erroHandler);
+    };
+    writable.on('error', erroHandler);
```
tankakatan, line 20: What if the `logstash` is unavailable, I guess we will fall into an infinite loop, won't we?

src/util/createUdpStream.ts
```diff
@@ -12,15 +15,29 @@ function createWritable(options: { address: string; port: number }) {
 }
 
 export function createUdpStream(options: { address: string; port: number }) {
+    let failures = 0;
     let writable = createWritable(options);
     const erroHandler = () => {
-        writable = createWritable(options);
-        writable.on('error', erroHandler);
+        if (failures < MAX_FAILURES) {
```
tankakatan, line 21: I could not find a stop condition here. The currently implemented retry logic looks like this:

_If the logstash is unavailable, recreate udp stream 10 times with no delay, then continue recreating it forever but with a delay of 10 secs_

IMO if the logstash is unavailable (or the internet connection is broken) it makes no sense to infinitely recreate the transport, and therefore I propose to implement it like this:

_if the logstash is unavailable, recreate udp stream 5 times with a delay of 5 seconds, and then if it still fails, **do not** recreate it again, just ignore it and continue writing logs into the log file_

tankakatan (approved):
Please update

tankakatan (approved):

PR #537, opened 2022-08-01
Add upload file script

Koochr:


Koochr:
I forgot I've already added `upload` command which does the same thing :)

PR #542, opened 2022-08-03
PN-200 Decode tx input data

germanviescas:
If the RPC request to `sendTransaction` includes information about which Smart Contract it wants to call, Point Engine will fetch the ABI, decode the transaction input data and added to the response, so that clients can display human-readable information to users.

In order not to introduce any breaking changes, if the contract information is not provided, the response is the same as it was before.

Also to maintain backwards compatibility, the decoded data is added as a new key in the payload.

### DIFF
--- a/src/rpc/decode.ts
+++ b/src/rpc/decode.ts
@@ -15,8 +15,48 @@ type DecodedTxInput = {
     params: Param[];
 };
 
-// A map to keep track of which ABIs have been added, to avoid duplicates.
-const addedABIs: Record<string, boolean> = {};
+/**
+ * Keep track of which ABIs have been added, to avoid duplicates, with a TTL.
+ */
+export class ABITracker {
+    private addedABIs: Record<string, number> = {};
+    private ttlMs: number;
+
+    constructor(ttlMs?: number) {
+        this.ttlMs = ttlMs ?? 10 * 60 * 1000; // default: 10 minutes.
+    }
+
+    private key(target: string, contract: string): string {
+        return `${target}:${contract}`;
+    }
+
+    private isExpired(key: string): boolean {
+        if (!this.addedABIs[key]) {
+            return false;
+        }
+        const ttl = this.addedABIs[key];
+        return Date.now() - ttl > this.ttlMs;
+    }
+
+    public has(target: string, contract: string): boolean {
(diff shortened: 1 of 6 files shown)


src/rpc/decode.ts
```diff
@@ -0,0 +1,46 @@
...
+    const identity = new URL(target).host.replace(/.point$/, '');
+    const {_jsonInterface} = await ethereum.loadWebsiteContract(identity, contract);
+
+    if (!addedABIs[`${target}:${contract}`]) {
+        abiDecoder.addABI(_jsonInterface);
+        addedABIs[`${target}:${contract}`] = true;
+    }
+
+    const decoded = abiDecoder.decodeMethod(txInputData);
+    return decoded ? (decoded as DecodedTxInput) : null;
```
tankakatan, line 45: What does faulty `decoded` variable mean? Should we consider throwing an error in this case. Or is it a normal scenario?

germanviescas: When `abiDecoder.decodeMethod` fails, it returns `undefined`, instead of for example throwing an error.

We could throw an error if we wanted. It comes down to how critical we think it is to not be able to decode the tx input.

My thinking was that it wasn't critical, if for any reason, we couldn't decode it. The confirmation window will show the raw data and users can decide if they want to confirm or not. But if we want to treat it as something critical that should stop the flow, we could do it.

There's the case of interacting with contracts that are not deployed to Point Network. In those cases, we won't be able to fetch the ABI (at least with these changes); so, decoding will not work. I'm trying to think if this is a valid use case though, I'm not sure if it's something plausible.

tankakatan: Thanks @germanviescas, of course, I agree it makes sense to let user decide whether they want to proceed with the request or not.

I just thought that internally we still want to know exactly what happens. The `decodeMethod` call may fail due to multiple reasons, either the ABI is outdated or the provided parameters are incorrect or maybe some other reason.

We can integrate the feature early and leave it as is for now, but I expect some issues here, and want to make sure we will be able to debug it.

At least let's log everything we know about the failure. Like, the contract and the method names, the provided args, etc etc.

While we'll be able to debug it, we can still let user make their own choice regarding the tx.

WDYT?

tankakatan (approved):

PR #556, opened 2022-08-09
Increase wallet timeout to wait for SNS/ENS

germanviescas:
We now look for SNS and ENS domains when resolving addresses. The additional requests sometimes cause timeouts.

Until we implement a better solution, like caching, I propose to increment the timeout to prevent users from getting this error.


germanviescas:
Closing this as there is now a timeout on Point Engine's side.

PR #566, opened 2022-08-11
Serge improvements aug 7

sergevar:
This is an update from Aug 7, so needs to be merged properly
- a few bugfixes to renderer (a bunch of returns and awaits were missing, so zhtml error messages weren't properly propagating to the browser), AR mint not reading the environment mode properly
- add throw tag into zhtml {% throw 'anything' %} will throw 'anything' as an error, instead of using a bunch of multi-nested ifs in zhtml logic
- two new web2 domain redirects with CORS enabled: fonts.googleapis.com -> gfonts.mirror.point, fonts.gstatic.com -> gfontfiles.mirror.point

While it's being merged, I'll try to upload the mirrors content and we can test whether we now have fully web3 copy of Google Fonts perfectly working

Upd: +if there is a .local domain, it tries to resolve it using zappsdir even in production mode. That was a quick hack for me while the dev env was (still is?) broken. Lmk your thoughts on whether any issues might arise out of it. Maybe we should disable it

### DIFF
--- a/src/client/proxy/handlers/common.ts
+++ b/src/client/proxy/handlers/common.ts
@@ -40,7 +40,7 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
 
         if (host === 'point') {
             if (req.url.startsWith('/web2redirect')) {
-                res.header('content-type', 'text/html');
+                res.header('Content-Type', 'text/html');
                 let refererHost = req.headers.referer || '';
                 const matches = refererHost.match(/^https:\/\/(.*)\//);
                 if (matches) {
@@ -68,7 +68,7 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
                 const file = await fs.readFile(filePath);
                 const contentType = ext ? getContentTypeFromExt(ext) : detectContentType(file);
 
-                res.header('content-type', contentType);
+                res.header('Content-Type', contentType);
                 return file;
             } else {
                 // If not, render index.zhtml
@@ -81,8 +81,8 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
             }
         } else if (host.endsWith('.z')) {
             res.header('Location', 'https://' + host.replace(/\.z/, '.point'));
-            res.status(301).send();
-        } else if (config.get('mode') === 'zappdev' && host.endsWith('.point')) {
+            return res.status(301).send();
+        } else if (host.endsWith('.local') || (config.get('mode') === 'zappdev' && host.endsWith('.point'))) {
             // when MODE=zappdev is set this site will be loaded directly from the local system - useful for Zapp developers :)
             // Side effect: versionig of zapps will not work for Zapp files in this env since files are loaded from local file system.
             const version = (queryParams.__point_version as string) ?? 'latest';
@@ -90,15 +90,15 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
             // First try route file (and check if this domain even exists)
             const zrouteId = await blockchain.getZRecord(host, version);
 
-            if (!zrouteId) {
-                res.status(404).send('Domain not found (Route file not specified for this domain)');
+            if (!zrouteId && !host.endsWith('.local')) {
+                return res.status(404).send('Domain not found (Route file not specified for this domain)');
             }
 
             const zappName = host.endsWith('dev') ? `${host.split('dev')[0]}.point` : host;
 
-            const zappsDir: string = config.get('zappsdir');
+            const zappsDir = String(config.get('zappsdir'));
             let zappDir: string;
-            if (zappsDir !== undefined && zappsDir !== '') {
+            if (zappsDir !== 'undefined' && zappsDir !== '' && zappsDir !== 'null') {
                 if (zappsDir.startsWith('/') || zappsDir.startsWith('~')) {
                     zappDir = path.resolve(zappsDir, zappName);
                 } else {
@@ -115,8 +115,8 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
                 rootDir = deployConfig.rootDir;
             }
 
-            const publicPath = path.resolve(zappDir, rootDir);
-            const routesJsonPath = path.resolve(zappDir, 'routes.json');
+            const publicPath = path.resolve(zappDir, rootDir); if (!FS.existsSync(publicPath)) throw new Error('Public path ' + publicPath + ' doesnt exist.');
+            const routesJsonPath = path.resolve(zappDir, 'routes.json'); if (!FS.existsSync(routesJsonPath)) throw new Error('Routes file ' + routesJsonPath + ' doesnt exist.');
             const routes = JSON.parse(await fs.readFile(routesJsonPath, 'utf8'));
 
             const {routeParams, templateFilename, rewritedPath} = getParamsAndTemplate(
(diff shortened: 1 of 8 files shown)


src/client/proxy/handlers/common.ts
```diff
@@ -110,8 +111,8 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
                 rootDir = deployConfig.rootDir;
             }
 
-            const publicPath = path.resolve(zappDir, rootDir);
-            const routesJsonPath = path.resolve(zappDir, 'routes.json');
+            const publicPath = path.resolve(zappDir, rootDir); if (!existsSync(publicPath)) throw new Error('Public path ' + publicPath + ' doesnt exist.');
```
Koochr, line 114: Purely stylish note, but anyway: I'm not sure why eslint doesn't show warning that the string is too long
We discussed some time ago that we should avoid `if` without braces (and also there are line breaks missing)

src/client/proxy/handlers/common.ts
```diff
@@ -133,25 +134,28 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
 
                 const renderer = new Renderer({localDir: publicPath} as any);
 
-                res.header('Content-Type', 'text/html');
+                const contentType = ext ? getContentTypeFromExt(ext) : 'text/html';
+                res.header('Content-Type', contentType);
+
                 // TODO: sanitize
-                return renderer.render(templateFilename, templateFileContents, host, {
+                return await renderer.render(templateFilename, templateFileContents, host, {
```
Koochr, line 141: `await` is redundant when returning from `async` function.
BTW I think we can add the rule for it: https://eslint.org/docs/latest/rules/no-return-await#:~:text=Using%20return%20await%20inside%20an,before%20resolving%20the%20outer%20Promise.

src/client/proxy/handlers/common.ts
```diff
@@ -198,9 +202,11 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
                 const templateFileContents = await getFile(templateFileId);
                 const renderer = new Renderer({rootDirId} as any);
 
-                res.header('Content-Type', 'text/html');
+                const contentType = ext ? getContentTypeFromExt(ext) : 'text/html';
```
Koochr, line 205: Can template file be something except HTML?

src/client/proxy/handlers/common.ts
```diff
@@ -198,9 +202,11 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
...
+                res.header('Content-Type', contentType);
+
                 // TODO: sanitize
-                return renderer.render(templateFileId, templateFileContents, host, {
+                return await renderer.render(templateFileId, templateFileContents, host, {
```
Koochr, line 209: Same as above regarding `await`

src/client/proxy/handlers/common.ts
```diff
@@ -225,13 +231,13 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
 
                         res.header('Content-Type', 'text/html');
                         // TODO: sanitize
-                        return renderer.render(templateFileId, templateFileContents, host, {
+                        return await renderer.render(templateFileId, templateFileContents, host, {
```
Koochr, line 234: Same

src/client/proxy/handlers/common.ts
```diff
@@ -350,13 +358,17 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
 
                         res.header('Content-Type', 'text/html');
                         // TODO: sanitize
-                        return renderer.render(templateFileId, templateFileContents, identity, {
-                            ...routeParams,
-                            ...queryParams,
-                            ...((req.body as Record<string, unknown>) ?? {})
-                        });
+
+                        return await renderer.render(
```
Koochr, line 362: Same with `await`

src/client/proxy/handlers/common.ts
```diff
@@ -409,11 +426,11 @@ const getHttpRequestHandler = () => async (req: FastifyRequest, res: FastifyRepl
                     });
                 }
             }
-            res.status(404).send('Not Found');
+            return res.status(404).send('Not Found');
         }
     } catch (e) {
         log.error({stack: e.stack, errorMessage: e.message}, 'Proxy error');
-        res.status(500).send('Internal engine error');
+        return res.status(500).send('Internal engine error: ' + JSON.stringify(e.message));
```
Koochr, line 433: Normally, server should not expose internal error to the client. However, our "server" is not a regular one, it's always local. But anyway, do you think it's needed? In case of development, we can always see the point node logs. In case of a bug, the point node logs go to logstash. But do regular users need these messages?

src/client/proxy/handlers/common.ts
sergevar: Actually, that was exactly the problem with error propagation.
Skipping await here leads to all errors from e.g. ZHTML renderer being thrown higher up instead of here on the spot, and they don't end up being caught with the local try...catch which returns the error message to the browser

src/client/proxy/handlers/common.ts
sergevar: https://github.com/pointnetwork/pointnetwork/pull/566#discussion_r943690899

src/client/proxy/handlers/common.ts
sergevar: https://github.com/pointnetwork/pointnetwork/pull/566#discussion_r943690899

src/client/proxy/handlers/common.ts
sergevar: https://github.com/pointnetwork/pointnetwork/pull/566#discussion_r943690899

src/client/proxy/handlers/common.ts
sergevar: Yes, in google font mirror I render CSS on the fly via zhtml, but then the browser refused to consume it due to content-type being text/html and not text/css

src/client/proxy/handlers/common.ts
sergevar: You can split the assignment from if here

Can you link to the discussion that we should avoid ifs without braces? Our code base is becoming a behemoth (this file is 471 lines) and I'm trying to compact it while refactoring, so splitting simple ifs into 3 lines where 1 would suffice seems like a waste

src/client/proxy/handlers/common.ts
sergevar: Three groups of people
- users don't care about errors
- we can look it up in logs
- but developers, when in dapp development mode, would be able to immediately see if e.g. zhtml template is faulty. It was very annoying to switch to console output each time.

Maybe later we can consult the mode and see if we're in dapp dev mode and only then show the error

But also consider the situations where users encounter an error in a dApp and complain to the dApp development team. In that case, it would be much better for them to send the full error instead of just "Internal error" (and users sending logs to external dapp dev teams can be privacy-breaking)

germanviescas (dismissed):

src/client/proxy/handlers/common.ts
Koochr: Sorry, I can't find the thread, but I remember we discusse it with @tankakatan and the team, and agreed on this.
This idea for writing 3 statements in 1 line still seems not good for code readabiliti IMO. And if we want to make files shorter, we could rather split the code into several files.

tankakatan (approved):

PR #567, opened 2022-08-11
Send empty `alias` when SNS/ENS times out

germanviescas:
Timeouts have been adding to the requests to resolve SNS/ENS domains. When an error occurs, we send the string `Error` as the alias, which causes the UI to render `Error` as the wallet address.

I think a better approach is to send an empty string. The UI is prepared to handle this, it will fallback to rendering the blockchain address.

Screenshot from 2022-08-11 10-24-13

### DIFF
--- a/src/api/controllers/WalletController.js
+++ b/src/api/controllers/WalletController.js
@@ -115,7 +115,7 @@ class WalletController extends PointSDKController {
                                         alias = '';
                                 }
                             } catch (e) {
-                                alias = 'Error';
+                                alias = '';
                             }
                         })()
                     ]);


Koochr:
Thanks @germanviescas, I also wanted to fix it, cool that you did it. But there are also some places in `WalletController` where `balance` is set to `Error` or the message of error, and this can cause UI to crash. Maybe you could fix these occurrences as well?

germanviescas:
The balance errors require a slightly different approach as there is no fallback. So I would prefer to keep this PR focused on address/alias and handle wallet errors separately please.

tankakatan (approved):

PR #568, opened 2022-08-11
Refactor proxy: collapse repeating code

sergevar:
`client/proxy/handles/common.ts` has 471 LOC, and a lot of the blocks are repeating, which could also lead to unsynchronized code between repeating blocks, and bugs.

This PR finally heavily refactors it by collapsing repeating code so it's easier to deal with.

`getHttpRequestHandler(...)` now only parses the request and prepares information for the routing, and then calls on `fulfillRequest(...)` which does the routing, zhtml rendering, static routing, and serves the response.

Not everything is tested (couldn't test .sol/.eth), so need your help with making sure it's not breaking anything, since it's a heavy refactor.

**Note:** I had to base it off of `serge_improvements_aug_7` branch https://github.com/pointnetwork/pointnetwork/pull/566 so it should only be merged after that PR.


germanviescas:
Something that doesn't seem to work is requesting a route of https://point other than index. For example, going to https://point and navigating to _/wallet_ in the client, works as expected, but directly going to https://point/wallet (like when refreshing the page) returns a 404.

Koochr:
> Something that doesn't seem to work is requesting a route of https://point other than index. For example, going to https://point and navigating to _/wallet_ in the client, works as expected, but directly going to https://point/wallet (like when refreshing the page) returns a 404.
> 
> The SPA handler must not be catching this request as there are two routes (`/` and `index.zhtml`), and then _/wallet_ is treated as a static asset, which of course is not found.

Yep, that's a tricky question, I also was looking at the bunch of repeating code in `common` handler, but though `if` branches are similar, each of them has slight differences so it's easy to break something.
My latest update on this file fixed this behavior for point explorer, but treating single-route SPA for point explorer and for other apps appeared different, as in the first case we load files from the FS, but in the other we download them from arweave. So the way how we check if a static file exists are completely different

sergevar:
@germanviescas you are right, I could reproduce and attempted to fix it with the latest release, please check.

There are two issues related to this that we will need to solve long term:

1. The hack for treating one route like a wildcard route. This is undocumented, and not expected/non-intuitive from developers standpoint.

```
    if (templateFilename) {
        return await tryFulfillZhtmlRequest(cfg, templateFilename, routeParams, host);
    } else {
        // This is a static asset, no routes matched
        try {
            return await tryFulfillStaticRequest(cfg, urlPath);
        } catch (e) {
            // TODO: Undocumented!
            // Handling the case with SPA reloaded on non-index page:
            // we have to return index file, but without changing the URL
            // to make client-side routing work
            if (e instanceof HttpNotFoundError && Object.keys(cfg.routes).length === 1) {
                ({templateFilename} = getParamsAndTemplate(cfg.routes, '/'));
                return await tryFulfillZhtmlRequest(cfg, templateFilename!, routeParams, host);
            } else {
                throw e;
            }
        }
    }
```

Long-term solution should be actually supporting wildcards and patterns in the routes.json file.
But I have another issue with route file syntax (`/smth/:varname` is very ugly, especially when you look at many routes like that, it's not immediately obvious to the brain that :varname should be treated like a variable, whereas something like `/smth/{varname}` is much better), so maybe we could fix all of these at once. Let's discuss in our threads.

2. The order in which we try things: the problem with having a catch-all route right now (`{"/[a-zA-Z0-9]*": "index.zhtml"}`) is that it would override static files as well, because we're asking for routes first.

Right now the routes always take priority. `if (templateFilename) {` above is basically querying a decision on whether the routes matched anything. Of course, if we have a catch-all route, it will always match everything and we never get to the static files.

Which is why I had to do this song and dance above with `tryFulfillStaticRequest` and `tryFulfillZhtmlRequest` in different order depending on the circumstances. In our SPA edge-case, if a route fails to match, it tries a static file first, and then goes into wildcard route mode.

A common practice I noticed in webserver configuration files like nginx is often to have the opposite: static files first, and only then go through the routes (which might contain catch-all routes), and only after that give up and throw 404.

Maybe we should also give the priority to static files then? This would assume we never need to override an ability for routes file to override static paths, because static files would always take priority.

--

Let's discuss in our threads, and please test the commit and see if it fixes everything.

PR #597, opened 2022-08-23
Update the blog deployment

Koochr:
Recreated it from the old PR as the git history seems to be messed up


src/client/proxy/handlers/storage.ts
```diff
@@ -8,21 +8,27 @@ import {Template, templateManager} from '../templateManager';
...
+        if (!req.headers['content-type']?.match('multipart/form-data')) {
+            return res.status(415).send('Only multipart/form-data is supported');
+        }
 
-            const file = await req.file();
-            if (!file) {
-                return res.status(400).send('No files in the body');
+        const files = [];
+        const fields: Record<string, any> = {};
+        for await (const part of req.parts()) {
```
tankakatan, line 18: This actually looks like a breaking change. I think `/_storage` path is used by our dapps and it is expected to accept files in the `file` field. @jensendarren could you please check that?

src/client/proxy/handlers/storage.ts
```diff
@@ -8,21 +8,27 @@ import {Template, templateManager} from '../templateManager';
 // TODO: we don't handle multiple files upload. But if we want to,
 // we should change the response format
 const attachStorageHandlers = (server: FastifyInstance) => {
-    ['/_storage/', '/_storage'].forEach(route => {
-        server.post(route, async (req, res) => {
-            if (!req.headers['content-type']?.match('multipart/form-data')) {
-                return res.status(415).send('Only multipart/form-data is supported');
-            }
+    server.post('/_storage/', async (req: any, res: any) => {
```
tankakatan, line 11: This change should be tested, as long as the path is used in many dapps like `social.point` and others. @jensendarren, could you please confirm it works like that?

src/network/deployer.ts
```diff
@@ -0,0 +1,270 @@
...
+import tar from 'tar-fs';
+import gunzip from 'gunzip-maybe';
+const log = logger.child({module: 'Deployer'});
+
+const IS_PACKAGED = Boolean((process as typeof process & {pkg?: unknown}).pkg);
+const PROXY_METADATA_KEY = 'zweb/contracts/proxy/metadata';
+
+// TODO: this is a temporary solution, supporting only limited number of dependencies
+const NPM_DEPS_URLS: Record<string, string> = {
+    '@openzeppelin/contracts': 'https://registry.npmjs.org/@openzeppelin/contracts/-/contracts-4.7.3.tgz',
```
tankakatan, line 22: I think the dependencies should also be read from blockchain or storage (maybe we should upload a sort of dependency list or a `package.json` file). Otherwise we'll have to release new Engine every time these dependencies change.

src/network/deployer.ts
```diff
@@ -0,0 +1,270 @@
...
+        artifactsStorageId,
+        version
+    );
+
+    return artifactsStorageId;
+};
+
+const getProxyMetadataFilePath = async () => {
+    const networkNames: Record<number, string> = {
+        1: 'mainnet',
```
tankakatan, line 105: This should be added to config instead. Ideally it should be joined with the existing chain configurations.

internal/explorer.point/src/pages/DeployBlog.jsx
```diff
@@ -2,47 +2,13 @@ import axios from 'axios';
...
 import { useState } from 'react';
 
-// TODO: move somewhere
-const delay = (ms) =>
-    new Promise((resolve) => {
-        setTimeout(resolve, ms);
-    });
-
-const BLOG_FACTORY_ADDRESS = '0x1Df9546CC23D05D99fE05c0b75278d3172afB8C3';
+const VERSION = '0.1';
```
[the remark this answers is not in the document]
Koochr: This is a temporary hardcode unless we have a better solution. Currently we don't store it in blockchain

src/client/zweb/deployer_old/index.js
```diff
@@ -1,7 +1,7 @@
 const path = require('path');
 const fs = require('fs');
 const logger = require('../../../core/log');
-const log = logger.child({module: 'Deployer'});
+const log = logger.child({module: 'Deployer_old'});
```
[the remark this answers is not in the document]
Koochr: What do you mean by "majority of cases"? The only place where this deployer is used is in a manual `deploy` command. My deployer is not a blog deployer, now it can be used only for upgradeable contract deployment, but the idea is that it will replace the old one over time.
I can do the opposite and rename the new deployer to `deployer_new` if it makes sense

hardhat/hardhat.config.js
```diff
@@ -58,7 +59,13 @@ module.exports = {
             }
         ]
     },
-    paths: {
+    paths: IS_PACKAGED ? {
+        root: path.join(os.homedir(), '.point', 'hardhat'),
```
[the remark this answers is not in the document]
Koochr: They should not. In a packaged app we always use one directory for that

src/client/proxy/handlers/storage.ts
Koochr: The problem is that this endpoint is not working well and often freezes on file uploads, though this change doesn't help, so maybe it will be better to keep it as it was, yes

src/network/deployer.ts
Koochr: This is taken from the old deployer, I just didn't touch it. Though maybe it's simpler to remove completely

src/network/deployer.ts
Koochr: Again, this is the temporary solution, ideally we should add a function, which resolves this link using npm package name, but the goal was to make it work ASAP

amherag (approved):

PR #602, opened 2022-08-25
Point explorer dapp updates without events merged

jensendarren:


tankakatan (changes requested):
Please remove the `internal/explorer.point/public/index.js` file from the PR, the `internal/explorer.point/public` is now in `.gitignore` and shouldn't be committed again.

tankakatan (approved):

PR #615, opened 2022-08-30
Implement auth token

Koochr:


tankakatan (commented):
What is the reason you replace `wouter` with `react-router-dom`? Also please turn off linter auto fixes, as long as it produces a lot of irrelevant changes that complicate reviews and increase the risk some crucial issues will remain unattended.

Koochr:
> What is the reason you replace `wouter` with `react-router-dom`? Also please turn off linter auto fixes, as long as it produces a lot of irrelevant changes that complicate reviews and increase the risk some crucial issues will remain unattended.

wouter doesn't support query: https://github.com/molefrog/wouter/issues/58

amherag (approved):

PR #617, opened 2022-08-31
Fix the Point Browser not opening web2 links

JatinBumbra:
- Fixed the web2 links not opening by commenting out a line of code (didn't knew if it might be useful later on so commented it)
- Removed redundant `point-contract-manager` package from the root `package.json` 
- Installed `cross-env` for cross platform compatibility in the root `package.json` and updated & tested `scripts`
- Fixed the `clear` script in `internal/explorer.point/package.json`. The earlier script would create a `-p` directory and fail to delete the `public` directory. Now it works fine - tested on Windows using git-bash. Please test on other systems as well.

### DIFF
--- a/internal/explorer.point/package.json
+++ b/internal/explorer.point/package.json
@@ -28,7 +28,7 @@
     "wouter": "^2.7.4"
   },
   "scripts": {
-    "clear": "mkdir -p \"./public\" && rm -rf \"./public/*\" \".cache\" \".parcel-cache\"",
+    "clear": "rm -rf \"./public/*\" \".cache\" \".parcel-cache\"",
     "start": "react-scripts start",
     "watch:docker": "./watch.docker.js",
     "prebuild": "npm ci",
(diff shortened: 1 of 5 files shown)


jensendarren (approved):

internal/explorer.point/package.json
```diff
@@ -28,7 +28,7 @@
     "wouter": "^2.7.4"
   },
   "scripts": {
-    "clear": "mkdir -p \"./public\" && rm -rf \"./public/*\" \".cache\" \".parcel-cache\"",
```
germanviescas, line 31: I assume you had to remove the `mkdir` call because it doesn't work on Windows, but unfortunately it breaks the tests, please see the error [here](https://github.com/pointnetwork/pointnetwork/runs/8112509486?check_suite_focus=true).

One solution could be to ignore errors from `rm` (like sending stderr to /dev/null), but not sure if that kind of redirection would work on Windows. Another solution could be to use a cross-platform tool for `mkdir -p`.

germanviescas: Actually, given the `build` script, in explorer.point, we need the `public/` directory to be created. So, unless we modify `build`, my first alternative solution above won't work.

jensendarren (approved):

PR #634, opened 2022-09-06
Comment out the logs which can be avoided

JatinBumbra:


tankakatan (changes requested):
Thanks @JatinBumbra, you've spotted very good cases.

Please address the comments I've made, and also take a look at [this](https://github.com/pointnetwork/pointnetwork/blob/develop/src/db/index.ts#L30) and [this](https://github.com/pointnetwork/pointnetwork/blob/develop/src/client/proxy/index.ts#L43) lines, where we also need to change the log level to `trace`.

Along with the changes you made it will make the output much cleaner.

tankakatan (approved):

PR #652, opened 2022-09-13
Set xnet as a default network for development

Koochr:
Without this change, point explorer is bound to `mainnet`, which is of no use for development.
Of course, it can be overridden by `local-development` config, but IMO it makes sense to use `xnet` by default in development.


amherag (approved):

amherag:
Makes sense to me.

tankakatan (approved):

PR #656, opened 2022-09-15
PZ-431

JatinBumbra:
Fix the web2 links not opening by removing the `protected` flag on the web2 API route. No need to now send the token along with open web2 link request.


jensendarren (approved):

jensendarren:
@tankakatan I've tested this locally and it works.

tankakatan (approved):

PR #657, opened 2022-09-15
Throw error when target identity is not registered

germanviescas:


tankakatan (approved):
Thanks @germanviescas, I have to first ask Serge to verify the general workflow.

PR #658, opened 2022-09-16
remove protection for encryptedStorage get

jensendarren:


Koochr:
Why do we need it? Encrypted storage should be protected  IMO

tankakatan:
> Why do we need it? Encrypted storage should be protected IMO

I agree with @Koochr. At first I was also going to remove protection from it. But then I checked closely and found out that this endpoint actually decrypts the users encrypted content with their private key, which IMO is not safe to keep unprotected.

PR #659, opened 2022-09-16
Hacky config fix

Koochr:
Without this hack, the default config picked up by the app is `development`. And it always was so, we just didn't mention.
Setting `NODE_ENV=production` doesn't help for packaged app.
So though this fix seems hacky, I think it's the best possible solution.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,10 @@ import {Command} from 'commander';
 import disclaimer from './disclaimer';
 import {getContractAddress, compileAndSaveContract} from './util/contract';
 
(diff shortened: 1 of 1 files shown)


tankakatan (changes requested):
Actually I think we should not "hack" the intended workflow, as long as IMO there's nothing bad in it. Let's keep on using the `development` env as we've been using it before, and create a new one for `xnet` instead. On the other hand hacking it may bring more issues along with the minor improvements.

Let's simply add `xnet.yaml` environment, and a new script `"start:xnet": "NODE_ENV=xnet npm start"` in the `package.json`.

PR #664, opened 2022-09-16
Set production as a default config for the packaged app

Koochr:


package.json
```diff
@@ -129,7 +129,7 @@
     "start:dev": "cross-env NODE_ENV=development npm run watch",
     "start:e2e": "cross-env MODE=e2e npm run watch",
     "watch": "npm run --prefix ./internal/explorer.point --ignore-scripts build && nodemon",
-    "build": "perl -i -pe\"s|timeout: 20000|timeout: 60000|\" node_modules/@trufflesuite/web3-provider-engine/subproviders/rpc.js && tsc",
+    "build": "npm run patch && tsc",
```
tankakatan, line 132: This patch is not needed here. It is already run in `poistinstall` script where it actually belongs. We don't need this patch to run on every build.

tankakatan (changes requested):
Looks good, please address a small comment in package.json.

tankakatan (approved):

PR #665, opened 2022-09-16
PN-223 sol to point

germanviescas:
Please refer to [the ticket](https://point-labs.atlassian.net/browse/PN-223) for context.

## READ

The `identityToOwner` response has a new `pointAddress` key.

- If the identity is _.point_, `pointAddress` will be the same as `owner`.
- if the identity is _.sol_, `pontAddress` will be the Point Address of the current user, and `owner` will be the Solana address that owns the .sol domain.
- if the identity is _.eth_, `pointAddress` will be the same as `owner`, and will refer to the Ethereum address that owns the .eth domain.

It was implemented like this to keep the `owner` key as it was, and avoid breaking changes.

## WRITE

The homepage of Point Explorer shows a recommendation to SOL domain owners to "link" their POINT address. In practice, this means writing the Point Address to the `POINT` record of the Solana Domain Registry.

If users accept the recommendation and click the "link" button, we use Point SDK to communicate with Point Engine and follow the regular flow, where the user is presented with a confirmation window to accept the transaction.

This is not mandatory, so users can ignore our recommendation and dismiss the message on Point Explorer's homepage.

For a quick demo, please check out [this video](https://drive.google.com/file/d/1xDcPzBUbEjGDeM90Q52LzauUyOp-fvjI/view?usp=sharing).

### DIFF
--- a/internal/explorer.point/src/components/LinkPointToSol.jsx
+++ b/internal/explorer.point/src/components/LinkPointToSol.jsx
@@ -1,0 +1,119 @@
+import { useState, useEffect } from 'react';
+import Swal from 'sweetalert2';
+import { useAppContext } from '../context/AppContext';
+
+const USER_REJECTION_CODE = 4001;
+const LS_KEY = 'point_to_sol_msg';
+const EMPTY_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+const styles = {
+    container: {
+        borderRadius: 4,
+        boxShadow: '0px 7px 15px 0 rgba(0, 0, 0, .15)',
+        padding: 16,
+        marginTop: 24,
+        color: 'rgb(124, 124, 124)',
+        display: 'grid',
+        gridTemplateColumns: '1fr 200px',
+        alignItems: 'center',
+        gridColumnGap: 42,
+    },
+    description: {
+        margin: 0,
+    },
+};
+
+const LinkPointToSol = () => {
+    const { walletIdentity } = useAppContext();
+    const [hasPointAddress, setHasPointAddress] = useState(true);
+    const [display, setDisplay] = useState(
+        () => localStorage.getItem(LS_KEY) !== 'hide',
+    );
+
+    useEffect(() => {
+        async function fetchData() {
+            try {
+                const { data } = await window.point.identity.identityToOwner({
+                    identity: walletIdentity,
+                });
+                if (!data.pointAddress || data.pointAddress === EMPTY_ADDRESS) {
+                    setHasPointAddress(false);
+                }
+            } catch (err) {
+                console.error(err);
+            }
+        }
+
+        if (
+            walletIdentity.endsWith('.sol') &&
+            window.point.point.link_point_to_sol
+        ) {
+            fetchData();
+        }
+    }, [walletIdentity]);
+
+    const linkPointToSol = async () => {
+        try {
+            await window.point.point.link_point_to_sol(walletIdentity);
+            await Swal.fire(
+                'Done!',
+                'You have successfully linked your POINT address to your SOL domain.',
+                'success',
+            );
+            setDisplay(false);
+        } catch (err) {
+            if (err.code === USER_REJECTION_CODE) {
+                await Swal.fire('Request cancelled.', '', 'info');
+            } else {
+                await Swal.fire(
+                    'Sorry, something went wrong',
+                    'Please try again later or contact support',
+                    'error',
+                );
+                console.error(err);
+            }
+        }
+    };
+
+    const handleRemindMeLater = () => {
+        setDisplay(false);
+    };
+
+    const handleDoNotShowAgain = () => {
+        localStorage.setItem(LS_KEY, 'hide');
+        setDisplay(false);
+    };
+
+    if (hasPointAddress || !display) {
+        return null;
+    }
+
+    return (
+        <div style={styles.container}>
+            <p style={styles.description}>
+                We've noticed you're using a <em>SOL</em> domain, for the best
+                experience, we recommend linking your <em>POINT</em> address to
+                your <em>SOL</em> domain.
+            </p>
+            <button className="btn btn-md btn-success" onClick={linkPointToSol}>
+                Link POINT to SOL
+            </button>
+            <div>
+                <button
+                    className="btn btn-sm btn-link"
+                    onClick={handleRemindMeLater}
+                >
+                    Remind me later
+                </button>
+                <button
+                    className="btn btn-sm btn-link"
+                    onClick={handleDoNotShowAgain}
+                >
+                    Don't show this again
+                </button>
+            </div>
+        </div>
+    );
+};
+
+export default LinkPointToSol;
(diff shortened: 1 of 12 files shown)


internal/explorer.point/src/components/LinkPointToSol.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    setHasPointAddress(false);
+                }
+            } catch (err) {
+                console.error(err);
+            }
+        }
+
+        if (
+            walletIdentity.endsWith('.sol') &&
+            window.point.point.link_point_to_sol
```
germanviescas, line 49: This is here to make sure that the version of Point SDK in use exposes this method. This way deployments of Point Engine and Point SDK are not coupled.

internal/explorer.point/src/components/LinkPointToSol.jsx
```diff
@@ -0,0 +1,119 @@
...
+        color: 'rgb(124, 124, 124)',
+        display: 'grid',
+        gridTemplateColumns: '1fr 200px',
+        alignItems: 'center',
+        gridColumnGap: 42,
+    },
+    description: {
+        margin: 0,
+    },
+};
```
germanviescas, line 24: I kept everything in this new file, including some styles, because I understand there's a big refactor of Point Explorer in progress, so this way I can avoid conflicts and adjust to whatever new design is implemented.

internal/explorer.point/src/components/LinkPointToSol.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+        if (
+            walletIdentity.endsWith('.sol') &&
+            window.point.point.link_point_to_sol
+        ) {
+            fetchData();
+        }
+    }, [walletIdentity]);
+
+    const linkPointToSol = async () => {
```
tankakatan, line 55: **Minor:** I think we could leverage `useCalback` here.

internal/explorer.point/src/components/LinkPointToSol.jsx
tankakatan, line 49: **Minor:** the `fetchData` function does not depend on `window.point.point.link_point_to_sol`, but the `linkPointToSol` does. Also when dealing with function I think it is safer to check it like this `typeof window.point.point.link_point_to_sol === 'function'`.

src/util/cookieString.ts
```diff
@@ -42,10 +42,13 @@ export function encodeCookieString(obj: Record<string, string>): string {
 export function merge(old: string, obj: Record<string, string>): Record<string, string> {
     const prevData = parseCookieString(old);
 
-    // Delete `point` keys to avoid inconsistencies like having a `pn_alias` together with a `pn_root`,
-    delete prevData.pn_root;
-    delete prevData.pn_routes;
-    delete prevData.pn_alias;
+    // If there's a pn_alias, there cannot be pn_root + pn_routes (and viceversa)
+    if (obj.pn_alias) {
```
tankakatan, line 46: Sorry I missed from the beginning. The function name is `merge` which makes it unclear it may omit some keys in the resulting object. Might be confusing for other devs. Let's add an intermediary function or rename this one to `mergeAndResolveRecordConflicts` or something like that.

tankakatan (approved):
Thanks @germanviescas, everything looks good in general. I just have a few suggestions regarding the UX:

* The `LinkPointToSol` component should be positioned absolutely, as it currently shifts the entire layout down and makes it hop back again every time a user hides the dialog.
* I propose to also add a `link` button to the `@identity` tab of the explorer and keep it there even if a user clicks `do not show again`. It may be useful in case the user clicks this button unintentionally.
* Beside `remind me later` and `do not show again` we should also have a `learn more` button which will either expand the component or open another page where a user could see a full list of the features they will enable by linking together their point address and `.sol` domain.

Also please see a few minor issues I mentioned below.

internal/explorer.point/src/components/LinkPointToSol.jsx
germanviescas: > Also we should notify user that the linking implies a fee.
The fact that it displays the Confirmation Window should be indication enough in my opinion. That's the same way users approve all fees. What it would be great to add, is estimated fees to the confirmation window.

I'll make the rest of the UI changes. I'll try to come up with some text for the "learn more" section, but if you have something in mind, please send it through.

internal/explorer.point/src/components/LinkPointToSol.jsx
germanviescas: I didn't want `fetchData` to run if `window.point.point.link_point_to_sol` is not defined, because it would a waste of a network request (including a blockchain call).

Checking for `undefined` worked well on my tests. But no problem, I'll change it to `typeof`.

germanviescas: Regarding `useCallback`, it is an unnecessary optimization in this case in my opinion. Defining a new function is a very cheap operation, and the only thing the function is doing is the data fetching. It is not used as a dependency anywhere else, and the component is in the homepage, which will not get many re-renders.

Generally speaking, I don't think it's a good idea to add `useCallback`, `useMemo` and other optimization techniques unless performance issues require it.

src/util/cookieString.ts
germanviescas: > Might be confusing for other devs.

Hopefully, the time I take to annotate functions with comments helps other developers.

germanviescas:
Position absolute doesn't look very nice because there's little room between the navbar and the title.

To me, the layout shift is not a problem, considering this is kind of a "one off". But 
@tankakatan , if you think that looks too bad, maybe I could position it at the bottom of the page? Or do you think it's fine for it to almost overlap with the title?

Screenshot from 2022-09-20 08-17-53

germanviescas:
This is what I have for "lean more":
Screenshot from 2022-09-20 08-53-39

tankakatan:
> Position absolute doesn't look very nice because there's little room between the navbar and the title.
> 
> To me, the layout shift is not a problem, considering this is kind of a "one off". But @tankakatan , if you think that looks too bad, maybe I could position it at the bottom of the page? Or do you think it's fine for it to almost overlap with the title?
> 

I'd recommend making the notification bar smaller and/or reduce the amount of text. For example instead of `Link Point to SOL`, it could be just `Link` (what exactly is going to be linked is already clear from the notification text). It will already make the text to the left fit single line. If it's not enough, we could also rephrase the notification message and make it a few words shorter. In order to support really small screens the notification could be sized relatively.

tankakatan:
> This is what I have for "learn more":

Great, looks good for now. Maybe let's just write `send POINTs` (I mean, in plural).

germanviescas:
Updated placement of the message:

Screenshot from 2022-09-20 09-35-12

src/util/cookieString.ts
tankakatan: @germanviescas I really appreciate the time you take to annotate the code you provide, it really helps in most of the cases.

But please consider my point of view. If I see a function name like `merge` which is a super-basic operation, it would never occur to me that I have to go and check its' source code. Just because I expect it to be something very simple.

Now think of it a bit further. Imagine one has to make a super-urgent hot fix that involves using this function. In this case most of the devs would probably think it must be a library function.

Most likely not only they won't check the source code, they even wouldn't bother testing its' result, which will of course cause the hot fix release to be buggy.

Therefore please don't get irritated by comments like that, I am just trying to cover some extreme and rare cases that still can sometimes happen.

tankakatan (approved):

PR #670, opened 2022-09-20
PN-226 Add Point Address to Identity page of SOL domains

germanviescas:
These changes add a row to the table that is displayed in the Identity page. This additional row is only displayed when dealing with a `.sol` identity, since the `owner` of such identities is a Solana address and not a Point one.

This is what it looks like in four different scenarios.

Owner seeing their own identity, before linking their Point address:
owner_no_point_addr

Owner seeing their own identity, after having linked their Point address:
owner_point_addr

Visitor seeing a `.sol` identity that does not have a Point address linked to it:
visitor_no_point_addr

Visitor seeing a `.sol` identity that has a Point address linked to it:
visitor_point_addr

### DIFF
--- a/internal/explorer.point/src/hooks/useLinkPointToSol.js
+++ b/internal/explorer.point/src/hooks/useLinkPointToSol.js
@@ -1,0 +1,40 @@
+import { useCallback } from 'react';
+import Swal from 'sweetalert2';
+
+const USER_REJECTION_CODE = 4001;
+
+const successMessage = `
+    <p>You have successfully linked your POINT address to your SOL domain.</p>
+    <p>
+        <small>
+            Please mind that changes may take a few minutes to take effect.
+        </small>
+    </p>
+`;
+
+/**
+ * Makes the request to set the Point Address in the `POINT` record of a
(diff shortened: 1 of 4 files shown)


germanviescas:
After successfully linking the address, I tried refreshing the page so that the user would see the address instead of the outdated "link your address" message. However, because it takes a few moments for changes to take effect, this didn't work. That's why I added a heads-up to the success message.

If we really want to avoid having users to refresh manually, I could implement an optimistic update, passing the address from Point Engine to the browser. Please let me know.

internal/explorer.point/src/hooks/useLinkPointToSol.js
```diff
@@ -0,0 +1,40 @@
+import { useCallback } from 'react';
+import Swal from 'sweetalert2';
+
+const USER_REJECTION_CODE = 4001;
+
+const successMessage = `
```
tankakatan, line 6: **Minor & For the future:** let's figure out is there a better way to customise the success dialog.

tankakatan (approved):

PR #684, opened 2022-09-26
Change content type detection way

Koochr:
The idea:
- In the static handler, we first detect content type, using the lib. If it's detected as `text/plain`, we look at the extension, to handle `css` and `js`. Previously, we were first relying on the extension, so the content-type could be spoofed by malicious website.
- In HTML handler I've removed detection by extension and always return `text/html` instead, as it doesn't make sense.
I've tested explorer, social, email, created blog and drive, everything is working.

### DIFF
--- a/src/client/proxy/handlers/common.ts
+++ b/src/client/proxy/handlers/common.ts
@@ -282,14 +282,18 @@ const tryFulfillZhtmlRequest = async (
         renderer = new Renderer({rootDirId: cfg.remoteRootDirId} as any);
     }
 
-    res.header('Content-Type', 'text/html');
-
     // TODO: sanitize
-    return await renderer.render(templateId, templateFileContents, host, {
+    const rendered = await renderer.render(templateId, templateFileContents, host, {
         ...routeParams,
         ...queryParams,
         ...((req.body as Record<string, unknown>) ?? {})
     });
+    const contentType = detectContentType(Buffer.from(rendered));
+    if (!(contentType.match('text/html'))) {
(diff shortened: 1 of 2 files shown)


src/client/proxy/handlers/common.ts
```diff
@@ -282,9 +282,7 @@ const tryFulfillZhtmlRequest = async (
         renderer = new Renderer({rootDirId: cfg.remoteRootDirId} as any);
     }
 
-    const contentTypeFromExt = getContentTypeFromExt(ext || '');
-    const contentType = contentTypeFromExt || 'text/html';
-    res.header('Content-Type', contentType);
+    res.header('Content-Type', 'text/html');
```
tankakatan, line 285: This assumption doesn't seem correct. Even though the function is called `tryFulfillZhtmlRequest`, in `templateFilename` argument it actually receives a route that matches the requested path. As long as we currently don't put any restrictions on values added in the `routes.json` file, a route may resolve to a file of any type, and the `tryFulfillStaticRequest` call will miss it. Therefore we still need to check the content type here, and probably prevent rendering if it is something other than `text/html`.

tankakatan (changes requested):

src/client/proxy/handlers/common.ts
Koochr: Wait, how can that happen? We are returning `renderer.render`, which has `Twig` under the hood. Also, we call `.toString()` then. Is there any chance it can render an image or a binary?

tankakatan (approved):

PR #695, opened 2022-09-30
PZ-436 - Cryptography documentation

michelnapo:


amherag (approved):

tankakatan (changes requested):
Thanks @michelnapo! It generally looks good. However I think we now have some problems with typescript. Please have a look [here](https://github.com/pointnetwork/pointnetwork/actions/runs/3160783204/jobs/5145627585#step:6:9). It looks like the build process is broken which will also block the release.

tankakatan (approved):
Thanks @michelnapo, it looks great now!

PR #697, opened 2022-10-02
Feature/pn 241

Koochr:
- Removed all real URLs from `default` config
- Fixed the tests accordingly
- It turned out that we need `solana` and `rinkeby` to make the ENS work, so I had to add them to `devlocal` (and probably will need to add also to e2e envs, but I think they not used for now)
- Added `NODE_CONFIG_ENV=production` to `npm start` script as we discussed in slack
- Also important note on all local configs: since `identity_contract_address` is now network-specific, the local network for every local config should always be called `local` (with the dynamic network name we won't be able to pass it using env var, and patch the identity contract address)
- After we release this update, we should not forget to check the gateway (it shouldn't be any problem there, but anyway it's config has changed)


.github/workflows/e2e-tests.yaml
```diff
@@ -16,7 +16,6 @@ jobs:
         node-version: 16.x
 
     - run: npm install
-    - run: npm run build
```
tankakatan, line 19: It would be better to run it explicitly rather than making it a part of test scripts. Please keep this step here and remove it from the `pretest` script.

package.json
```diff
@@ -121,16 +121,18 @@
     "typescript": "^4.7.3"
   },
   "scripts": {
-    "test": "NODE_CONFIG_ENV=test MODE=test jest --forceExit",
-    "test:ci": "cp ./resources/blockchain-test-key.json ./tests/resources/key.json && POINT_KEYSTORE=./tests/resources MODE=test jest --forceExit --reporters=default --reporters=jest-junit",
+    "pretest": "mkdir -p ./tests/resources/datadir/keystore && cp ./resources/blockchain-test-key.json ./tests/resources/datadir/keystore/key.json && npm run build",
+    "test": "npm run pretest && npm run build:explorer:dev && POINT_DATADIR=./tests/resources/datadir POINT_KEYSTORE=./tests/resources/datadir/keystore NODE_CONFIG_ENV=test MODE=test jest --forceExit",
```
tankakatan, line 125: The `npm run pretest` is not intended to be run manually, and may actually cause errors. It's a [npm feature](https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts) and will run automatically.

tankakatan (changes requested):
LGTM. Please address the below comments, and then it will be merged.

PR #707, opened 2022-10-11
PN-235 Prevent XSS in SVG

germanviescas:
SVGs are sanitized before being returned. The sanitation process consists of:

- removing any `<script>` tags
- remoing "script attributes" such as `onload`, which can also be used to execute JS

To play around with SVGs and see how they render in different situations, please visit _https://vdnamreg.point/xss.html_.

---

I think sanitizing is the best approach, however, I also want to mention an alternative: using CSP headers.

When an SVG is requested directly (not inside an ``), we could disallow `<script>` tags with a Content-Security-Policy. Such policy would break other things (please refer to the console errors in the screenshot below), however, these errors are probably not a big concern when rendering an image in the browser.

Screenshot from 2022-10-11 11-13-09


src/client/proxy/handlers/storage.ts
```diff
@@ -2,9 +2,11 @@ import {FastifyInstance, FastifyRequest} from 'fastify';
 const {uploadFile, getFile, FILE_TYPE} = require('../../storage');
 // @ts-expect-error no types for package
 import detectContentType from 'detect-content-type';
+import isSVG from 'is-svg';
```
germanviescas, line 5: I was trying to avoid new dependencies like this one. But when getting content from storage, where we don't have file extensions available, checking if something is an SVG has some corner-cases that I think justify using a third-party library. I went through the test cases in `is-svg` and they look like a good fit for our use-case.

`detectContentType` unfortunately does not properly recognize SVG, it returns _text/xml_ or _text/plain_.

src/util/sanitizeSVG.ts
```diff
@@ -0,0 +1,36 @@
...
+const {JSDOM} = jsdom;
+const dom = new JSDOM();
+
+export const sanitizeSVG = (content: Buffer | string): string => {
+    const original = Buffer.isBuffer(content) ? content.toString('utf-8') : content;
+    if (!isSVG(original)) {
+        throw new Error('Not a valid SVG.');
+    }
+
+    const div = dom.window.document.createElement('div');
```
germanviescas, line 13: The most robust approach I found is to use `jsdom` to actually render the SVG in this safe environment and use DOM APIs to find and remove malicious data.

Alternatively, we could use regexp and string parsing to sanitize the SVG without the additional overhead of `jsdom`. I think the approach with `jsdom` is more straight-forward and less error-prone. But please let me know if you think otherwise or think `jsdom` is going to make it much slower (I haven't noticed significant changes).

PR #709, opened 2022-10-11
Add file existence check to storage; Slightly improve the ABI check

Koochr:


src/network/providers/ethereum.js
```diff
@@ -155,7 +155,8 @@ const isIdentityAbiRelevant = async () => {
     const abiPath = path.resolve(CONTRACT_BUILD_DIR, 'Identity.json');
 
     try {
-        const abiAndMetadata = JSON.parse(await fs.readFile(abiPath));
+        if (!existsSync(abiPath)) return false;
```
tankakatan, line 158: Please remove this line. We need to know exactly what happened to the ABI file as long as this part may still contain bugs. Basically if the file doesn't exist, we'll get to the same catch block below where the error will be logged. It's not actually an error, so you may decrease the log level to `warn` instead, but the message itself should remain in the log.

src/network/providers/ethereum.js
```diff
@@ -200,7 +200,7 @@ const fetchAndSaveIdentityAbiFromStorage = async () => {
 
         log.debug('Successfully fetched identity contract abi from storage');
 
-        return (abisByContractName['Identity'] = abiAndMetadata);
+        abisByContractName['Identity'] = abiAndMetadata;
```
tankakatan, line 203: Please don't remove the `return` statement here.

tankakatan (changes requested):

src/network/providers/ethereum.js
Koochr: Why do we need to return assignment? It makes no sense and against the best practices

tankakatan (approved):

PR #719, opened 2022-10-13
HOTFIX: hide LinkPointToSol banner in home

jensendarren:


germanviescas:
Hi @jensendarren, could you please clarify the reason behind removing this banner?

I'm just asking because we have the same functionality [here](https://github.com/pointnetwork/pointnetwork/blob/develop/internal/explorer.point/src/components/PointAddressRow.jsx), and it's also mentioned [here](https://github.com/pointnetwork/pointnetwork/blob/develop/internal/explorer.point/src/components/PublicKeyRow.jsx). So, is the goal is to get rid of it completely, we should also take care of that. But if the goal is just to make it less prominent by removing it only from the homepage, then no further changes would be required.

tankakatan (approved):

PR #723, opened 2022-10-17
PN-262 Add metadata to decoded tx input

germanviescas:
These changes add some metadata to the transaction inputs, which is attached to the response that Point Engine sends when a transaction is stored for user confirmation.

The caller,  Point SDK, can use this additional metadata to enrich the information presented to the user in the confirmation window.

The metadata these changes add is related to inputs that look like storage IDs and empty hashes (_0x000..._).

- For an empty hash, it will simply return a "no_content" message.
- If the hash refers to a file on disk, a message indicating that it's a storage ID is returned.
- If the hash refers to a blockchain transaction hash for which a receipt can be fetched successfully, a message indicating it's a tx hash is returned.
- If the two checks above fail, a "not found" message is returned.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -69,4 +69,4 @@ name_services:
 payments:
   token_address: "0xbf9be54Df2001E6Bd044cED0E508d936A9d38b1D"
 rpc:
-  send_transaction_timeout_secs: 120
+  send_transaction_timeout_secs: 300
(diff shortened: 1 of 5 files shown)


config/default.yaml
```diff
@@ -69,4 +69,4 @@ name_services:
 payments:
   token_address: "0xbf9be54Df2001E6Bd044cED0E508d936A9d38b1D"
 rpc:
-  send_transaction_timeout_secs: 120
+  send_transaction_timeout_secs: 300
```
germanviescas, line 72: Since the confirmation window presents more information now, including links to storage, I think we should give users more time to review it. Hence, I've increased the timeout to confirm transactions from 2 to 5 minutes.

tankakatan (approved):

PR #725, opened 2022-10-18
Fix deployment of upgradable contracts

Koochr:
- Hardhat root dir moved to `datadir` unconditionally
- Upgraded `@openzeppelin/hardhat-upgrades`
- Replaced sync `fs` calls in `deployer` with async
- Fixed `.openzeppelin` error in CLI deployer by replacing `hre.upgrades.deployProxy` with patched `deployProxy` from the blog deployer
- Changed the way of compiling upgradable contracts in CLI deployer: now we symlink project contracts and dependencies instead of copying anything
- Removed version arg in CLI deployer (same as it was done in the blog deployer before), which led to a long events scanning
- Added `path.resolve` to CLI deployer - it was failing to work with relative paths in the packaged mode
- Commented out `upgradeProxy` code in CLI deployer, as it's not working anyway


src/client/zweb/deployer/index.js
```diff
@@ -427,13 +491,12 @@ class Deployer {
...
                             if (
-                                proxyAddress == null || 
-                                proxyDescriptionFileId == null || 
+                                !proxyAddress ||
+                                !proxyDescriptionFileId ||
                                 force_deploy_proxy 
                             ) {
                                 //deploy a new proxy
                                 log.debug('deployProxy call');
-                                const cfg = {kind: 'uups'};
```
tankakatan, line 436: What is the reason for removing this config?

tankakatan (changes requested):
Thanks for the effort, I think we should improve it even more. Installing `hardhat` as well as any other dependencies inside the dapp project doesn't seem to be a right move at all.

I think we should try symlinks instead. Symlinks are cross platform, [supported](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback) natively in Node.js and hence look very much reliable. We should try it.

src/client/zweb/deployer/index.js
Koochr: It's not used

tankakatan: @Koochr it was used before your change.

tankakatan (approved):

PR #728, opened 2022-10-20
Add cosmos client as a provider

brianunlam:


config/production.yaml
```diff
@@ -57,3 +57,9 @@ network:
       currency_name: "Sol"
       currency_code: "SOL"
       tls: true
+    cosmos_testnet:
+      type: 'cosmos'
+      name: 'cosmos_testnet'
+      http_address: 'cosmos-testnet-rpc.allthatnode.com:26657'
```
tankakatan, line 63: Two things here:
1. If we keep this config in production environment, it will appear in the Wallet page of the Point Explorer app in production, which is not a good thing to have unless we fully integrate with our Cosmos endpoint. Therefore please move the entire `cosmos_testnet` section to `development.yaml` for now.
2. It would be better to integrate with our own node right away. As long as test net is down, for simple read operations we can use `rpc-mainnet-1.point.space:26657` for now.

tankakatan (changes requested):
Everything's good, please fix all the linter errors detected and address other comments.

PR #750, opened 2022-10-31
Request confirmation for keplr's sendTx

germanviescas:
These changes introduce the basic flow to _intercept_ some calls to the Keplr client and trigger the confirmation window flow.

I have also made a small refactor so that the implementation is a bit more aligned with Ethereum's and Solana's.
It's just a suggestion because the Keplr integration (or maybe I should say the Cosmos integration) is different, so it's not clear to me if it makes sense to try to align to Ethereum and Solana. For example, if a lot of methods are going to require user confirmation and payloads are too different, this approach could be more verbose. But for now, I think it helps.

### DIFF
--- a/src/rpc/rpc-handlers.ts
+++ b/src/rpc/rpc-handlers.ts
@@ -6,8 +6,7 @@ import solana, {SolanaSendFundsParams, TransactionJSON} from '../network/provide
 import {decodeTxInputData, DecodedTxInput, addMetadata} from './decode';
 import {getNetworkPublicKey} from '../wallet/keystore';
 import logger from '../core/log';
-import {keplr} from '../network/providers/keplr';
-import {Keplr} from '@keplr-wallet/provider';
+import {send as keplrSend} from '../network/providers/keplr';
 
 const log = logger.child({module: 'RPC'});
 
@@ -102,6 +101,9 @@ const confirmTransaction: HandlerFunc = async data => {
                     );
                 }
                 break;
+            case 'cosmos':
+                result = await keplrSend({method: 'keplr_sendTx', ...tx});
+                break;
             default:
                 return {
                     status: 400,
@@ -181,30 +183,9 @@ const specialHandlers: Record<string, HandlerFunc> = {
             return {status: statusCode, result: {code: statusCode, message: err.message}};
         }
     },
-    keplr: async (data) => {
-        const {method, params} = data;
-        const methodName = method.split('_')[1];
-        // eslint-disable-next-line @typescript-eslint/ban-types
-        const keplrMethod = keplr[methodName as keyof Keplr] as Function;
-        if (typeof keplrMethod === 'function') {
-            try {
-                const result = await keplrMethod.apply(keplr, params) || {};
-                return {status: 200, result};
-            } catch (error) {
-                return {
-                    status: 400, result: {
-                        __error: {
-                            message: error.message,
-                            module: error.module,
-                            code: error.code
-                        }
-                    }
-                };
-            }
-
-        }
-        return {status: 500, result: 'Not valid method'};
-    }
+    // Keplr
+    keplr_sendTx: storeTransaction,
+    keplr_confirmTx: confirmTransaction
 };
 
 // Handlers for methods related to permissions.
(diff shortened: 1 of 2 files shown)


src/rpc/rpc-handlers.ts
```diff
@@ -181,30 +183,9 @@ const specialHandlers: Record<string, HandlerFunc> = {
...
-                    }
-                };
-            }
-
-        }
-        return {status: 500, result: 'Not valid method'};
-    }
+    // Keplr
+    keplr_sendTx: storeTransaction,
+    keplr_confirmTx: confirmTransaction
```
germanviescas, line 188: Here we _intercept_ the transactions that cannot be immediately sent to the blockchain. For now it's only `sendTx` but we might need to add others depending on whether they need user confirmation or not.

`keplr_confirmTx` is a made up tx name that we can use to detect user confirmation. In reality, this is not strictly needed as all confirmations are coming in as `eth_confirmTransaction` and then a network check is done in the `confirmTransaction` handler. I think this can be improved and we shouldn't have done it like this when Solana support was added, but it is out of scope for keplr integration and I suggest not to worry about it now.

src/rpc/rpc-handlers.ts
```diff
@@ -276,10 +257,6 @@ const handleRPC: HandlerFunc = async data => {
             const res = await permissionHandler({id, method, params, origin, network});
             return res;
         }
-        if (method.startsWith('keplr_')) {
-            const specialHandler = specialHandlers['keplr'];
-            return specialHandler({id, method, params, network, target, contract});
-        }
```
germanviescas, line 282: Instead of treating all `keplr_...` methods as special, we could explicitly list the _special_ ones in the `specialHandlers` as I've done above, and let the rest be handled by the `send` method I mentioned in a previous comment.

By _special_ in this context we mean methods that require user confirmation, and that means methods that write to the blockchain, hence requiring users to spend funds.

With this change, the flow is the same for the three providers:
* a request comes in
* we check if the method has a _specialHandler_
* if it does, we call the _specialHandler_
* if it doesn't, we call the `send` method for the given provider and immediately send the tx to the blockchain

src/rpc/rpc-handlers.ts
```diff
@@ -298,6 +275,9 @@ const handleRPC: HandlerFunc = async data => {
             case 'solana':
                 result = await solana.send({method, params, id, network});
                 break;
+            case 'cosmos':
+                result = await keplrSend({method, params, id, network});
+                break;
```
germanviescas, line 280: Following the comment above, the idea is that, if we end up here, the method is not _special_, meaning it does not require user confirmation. So, it can be sent to the blockchain, in this case, using a keplr client to do so.

germanviescas:
Let me share a test case.

In Point SDK, I added a `sendTx` method to `window.keplr`. In _src/pointsdk/keplrProvider.ts_ I added:

```
sendTx: async function () {
    return handleRequest({
        method: 'keplr_sendTx',
        params: [{name: 'testing', value: '47'}] // hardcoded for testing
    });
}
```

In Point Browser, I executed `window.keplr.sendTx()` from the console.

Point Engine (running this branch), got the request and the `ZProxySocketController` passed it to `handleRPC`.

`handleRPC` checked if `keplr_sendTx` was a special method. It was, so it called `storeTransaction`, which saved the transaction for later processing and sent a response to Point SDK that caused the confirmation window to be displayed:

sendTx


If `keplr_sendTx` hadn't been listed as a special method, the request would have been passed directly to keplr's `send` method (the one added in _src/network/providers/keplr.ts_).

If the user clicks CANCEL or closes the confirmation window, that's the end of it.

I clicked ALLOW, which sends a message to Point Engine that ends up in the `confirmTransaction` handler (in _src/rpc/rpc-handlers.ts_). `confirmTransaction` retrieves the tx that `storeTransaction` had saved before, and submits it to keplr's `send` method, so that it can finally be broadcasted to the blockchain.

For this test, I actually commented it out, and instead returned a mocked response from the `send` method. Which was sent to Point SDK and that was the end of it.

This is the mocked reponse from keplr's `send`, in Point SDK console.log:

confirmTx

tankakatan (approved):

src/rpc/rpc-handlers.ts
brianunlam: is not sendTx which requires confirmation but signTx

src/rpc/rpc-handlers.ts
brianunlam: this handler handles all the cases, it's not using rpc so old structure has no sense

PR #752, opened 2022-11-02
PN-287 add point to keplr chains

germanviescas:
Adding Point Network to the chains known by Keplr so that we can `enable` it from the client and use it.

Having Point Network as one of Keplr's chains, its `chainId` can be used in the client to interact with it. In the image below, I am enabling it, getting an offline signer, getting the account, and signing an empty object:

console

You may notice that the return value of the `signer.sign()` call in the screenshot above is `{ msg: "WIP" }`. That's hardcoded on the Engine side. In the final implementation, that response should actually trigger the confirmation window, but for now it's just logging the data and returning a dummy response, so that we can inspect it.

This is the message that the Engine gets when `signer.sign()` is called:
sign_message

### DIFF
--- a/src/rpc/rpc-handlers.ts
+++ b/src/rpc/rpc-handlers.ts
@@ -184,8 +184,10 @@ const specialHandlers: Record<string, HandlerFunc> = {
         }
     },
     // Keplr
-    keplr_sendTx: storeTransaction,
-    keplr_confirmTx: confirmTransaction
+    keplr_signAmino: async data => {
+        console.log(data);
+        return {status: 200, result: {msg: 'WIP'}};
+    }
 };
 
 // Handlers for methods related to permissions.
@@ -275,7 +277,9 @@ const handleRPC: HandlerFunc = async data => {
             case 'solana':
                 result = await solana.send({method, params, id, network});
                 break;
+            case 'cosmos':
             case 'keplr':
+                console.log({method, params});
                 result = await keplrSend({method, params, id, network});
                 break;
             default:
--- a/src/network/config.ts
+++ b/src/network/config.ts
@@ -106,7 +106,11 @@ import {
     EVMOS_RPC_ENDPOINT,
     EVMOS_RPC_CONFIG,
     EVMOS_REST_ENDPOINT,
-    EVMOS_REST_CONFIG
+    EVMOS_REST_CONFIG,
+    POINT_RPC_ENDPOINT,
+    POINT_RPC_CONFIG,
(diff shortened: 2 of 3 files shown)


src/rpc/rpc-handlers.ts
```diff
@@ -275,7 +277,9 @@ const handleRPC: HandlerFunc = async data => {
             case 'solana':
                 result = await solana.send({method, params, id, network});
                 break;
+            case 'cosmos':
             case 'keplr':
```
germanviescas, line 281: I had to add `cosmos` as an option here because the network in the request from Point SDK is actually `cosmos_testnet` (which has `type: cosmos`).

I would suggest we remove `keplr` here, but didn't want to do it just yet, as it will probably break when running in non-development mode, since `cosmos_testnet` only exists in the development config. I created ticket [PN-286](https://point-labs.atlassian.net/browse/PN-286) to deal with this.

germanviescas:
With Point Network added to keplr, and the little testing I could do directly from the console. I thought that maybe ping-pub would work. But unfortunately it did not.

On the wallet page, there is an option to enable keplr. As far as I understand, that option is there to add Point Network Chain configuration to Keplr. This is necessary in web2, but should not be necessary in Point Browser as that's exactly what we're doing in this PR, adding Point Network Chain to Keplr.

I clicked it nevertheless, to see what happens. I get the error: `window.keplr.experimentalSuggestChain is not a function`. It's a method that we haven't added on our `window.keplr` yet. I didn't worry too much about this because I don't think it's needed.

I selected Keplr as a device and tried to enable it, here:
Screenshot from 2022-11-02 15-40-25

It doesn't seem to work, it doesn't take me to the next step, but I see no errors. The only visual feedback is that the "1 Device" in the screenshot above turns red.

In Point Engine, I see the same requests that I see when I try directly from the console. The first two lines in the screenshot below are logged when I click the "Next >" button. The last two lines are logged when I manually call `enable` and `getOfflineSigner` from the console:
pingpub-vs-console

They are exactly the same, so my initial thought is that ping-pub might be expecting a different response for the `getOfflineSigner` method, but I don't know. I'm currently looking into this.

germanviescas:
By making `getOfflineSigner` sync, I was able to move to the next step:
Screenshot from 2022-11-03 09-13-07

germanviescas:
With this PR, plus [these changes in Point SDK](https://github.com/pointnetwork/pointsdk/pull/87), I've finally been able to successfully import a wallet.
Screenshot from 2022-11-03 12-11-51

PR #759, opened 2022-11-14
Add deploy/upgrade blog endpoint

brianunlam:


src/client/proxy/handlers/deployBlog.ts
```diff
@@ -0,0 +1,166 @@
...
+        await deployUpgradableContracts({
+            contracts: [{file: Buffer.from(blogContract), name: 'Blog'}],
+            version: VERSION,
+            target: subidentity,
+            forceDeployProxy: isNewIdentity,
+            dependencies: BLOG_DEPENDENCIES
+        });
+    }
+    if (rootDir !== ROOT_DIR_ID) {
+        await setRootDir(subidentity, ROOT_DIR_ID);
```
tankakatan, line 115: It would be much more readable if you put the code of this function here, as long as this function is just an alias of another one, and actually is not called anywhere else.

src/client/proxy/handlers/deployBlog.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (routes !== ROUTES_FILE_ID) {
+        await setRoutes(subidentity, ROUTES_FILE_ID);
```
tankakatan, line 118: It would be much more readable if you put the code of this function here, as long as this function is just an alias of another one, and actually is not called anywhere else.

tankakatan (changes requested):

